PR #5, opened 2021-04-08
added clangd server

ecmma:
This adds a basic clangd-11.0.0 server configuration for c, cpp and obj-c source files. 
I reckon there's some inner configuration that takes place thanks to `nvim-lspconfig` itself, as stated in the [documentation](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#clangd). 

Tested on Linux 5.11.10-arch1-1, but it should work fine for macOS as well. 
If I got anything wrong, I'm happy to fix it!

### DIFF
--- a/lua/nvim-lsp-installer/servers/clangd.lua
+++ b/lua/nvim-lsp-installer/servers/clangd.lua
@@ -3,8 +3,6 @@ local server = require('nvim-lsp-installer.server')
 local root_dir = server.get_server_root_path('c-family')
 
 local install_cmd = [=[
-rm -r clangd;
-
 if [[ $(uname) == Linux ]]; then
   wget -O clangd.zip https://github.com/clangd/clangd/releases/download/11.0.0/clangd-linux-11.0.0.zip;
 elif [[ $(uname) == Darwin ]]; then 
--- a/lua/nvim-lsp-installer/servers/clangd.lua
+++ b/lua/nvim-lsp-installer/servers/clangd.lua
@@ -1,0 +1,30 @@
+local server = require('nvim-lsp-installer.server')
+
+local root_dir = server.get_server_root_path('c-family')
+
+local install_cmd = [=[
+rm -r clangd;
+
+if [[ $(uname) == Linux ]]; then
+  wget -O clangd.zip https://github.com/clangd/clangd/releases/download/11.0.0/clangd-linux-11.0.0.zip;
+elif [[ $(uname) == Darwin ]]; then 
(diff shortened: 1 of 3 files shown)


lua/nvim-lsp-installer/servers/clangd.lua
```diff
@@ -0,0 +1,30 @@
...
+mv clangd_11.0.0 clangd;
+
+]=]
+
+return server.Server:new {
+  name = "clangd",
+  root_dir = root_dir, 
+  install_cmd = install_cmd, 
+  default_options = {
+    cmd = { root_dir .. '/clangd/bin/clangd'},
```
williamboman, line 28: It seems like [LSP config](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/clangd.lua#L30) provides a `--background-index` argument as well. I can't seem to find any docs or traces of it in clangd itself though, do you happen to know what it could be?

lua/nvim-lsp-installer/servers/clangd.lua
```diff
@@ -0,0 +1,30 @@
+local server = require('nvim-lsp-installer.server')
+
+local root_dir = server.get_server_root_path('c-family')
+
+local install_cmd = [=[
+rm -r clangd;
```
williamboman, line 6: You could omit this actually, when installing a new server the entire `root_dir` is [wiped](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/server.lua#L133-L135), so every install cmd is executed in a completely pristine state

williamboman (approved):
Thanks for the contribution ⭐️, LGTM!

lua/nvim-lsp-installer/servers/clangd.lua
ecmma: from `clangd --help`: 
`  --background-index              - Index project code in the background and persist index on disk.`

lua/nvim-lsp-installer/servers/clangd.lua
ecmma: removed with 24806b21d4e142790eb0b7b4f6ffc9cd5d631571 
Thank you!

williamboman:
ty @ecmma ❤️

lua/nvim-lsp-installer/servers/clangd.lua
mortezadadgar: Why this argument has not been added to default configuration, sounds useful

williamboman: Oh, had someone asked me I'd have thought it was added already. Happy to receive PR that adds it!

PR #8, opened 2021-04-14
added gopls server

ecmma:
This adds a basic gopls server configuration for Go source files. 
Since installing gopls requires the Go CLI, I thought that it would be better to avoid installing the CLI itself here, and it just aborts if Go isn't found in the system. Also, I don't think there's much control over the installation of the server itself, so `install_cmd` just checks if `gopls` exists afterwards. 

Tested on Linux 5.11.10-arch1-1, but it should work fine for other Go-supported systems as well.
If I got anything wrong, I'm happy to fix it!
(hello again! :))

### DIFF
--- a/lua/nvim-lsp-installer/servers/gopls.lua
+++ b/lua/nvim-lsp-installer/servers/gopls.lua
@@ -5,7 +5,7 @@ local root_dir = server.get_server_root_path('go')
 local install_cmd = [=[
 
 GO111MODULE=on GOBIN="$PWD" GOPATH="$PWD"  go get golang.org/x/tools/gopls@latest;
-! command -v ./gopls &> /dev/null; 
+command -v ./gopls &> /dev/null; 
 
 ]=]
 
--- a/lua/nvim-lsp-installer/servers/gopls.lua
+++ b/lua/nvim-lsp-installer/servers/gopls.lua
@@ -3,14 +3,9 @@ local server = require('nvim-lsp-installer.server')
 local root_dir = server.get_server_root_path('go')
 
 local install_cmd = [=[
-GO111MODULE=on GOBIN="$PWD" GOPATH="$PWD"  go get golang.org/x/tools/gopls@latest;
 
-if ! command -v ./gopls &> /dev/null; 
-then 
-  echo "something went wrong!"
-else 
-  echo "installation completed"
-fi
+GO111MODULE=on GOBIN="$PWD" GOPATH="$PWD"  go get golang.org/x/tools/gopls@latest;
+! command -v ./gopls &> /dev/null; 
 
 ]=]
 
@@ -24,6 +19,6 @@ return server.Server:new {
   end,
   install_cmd = install_cmd,
   default_options = {
-    cmd = {root_dir .. "/gopls"},
+    cmd = {root_dir .. "/gopls", "-logfile=/home/ecmm/log"},
   }
 }
--- a/lua/nvim-lsp-installer/servers/gopls.lua
+++ b/lua/nvim-lsp-installer/servers/gopls.lua
@@ -1,0 +1,31 @@
+local server = require('nvim-lsp-installer.server')
+
+local root_dir = server.get_server_root_path('go')
+
+local install_cmd = [=[
+if ! command -v go &> /dev/null;
+then 
+  echo "Please install the Go CLI before installing gopls.";
+  echo "refer to https://golang.org/doc/install";
+  exit 1;
+fi
+
+GO111MODULE=on go get golang.org/x/tools/gopls@latest;
+
+if ! command -v gopls &> /dev/null; 
+then 
+  echo "something went wrong!"
+else 
+  echo "installation completed"
+fi
+
+]=]
+
+return server.Server:new {
+  name = "gopls",
+  root_dir = root_dir,
+  install_cmd = install_cmd,
+  default_options = {
+    cmd = {"gopls"},
+  }
+}
--- a/lua/nvim-lsp-installer/servers/gopls.lua
+++ b/lua/nvim-lsp-installer/servers/gopls.lua
@@ -3,13 +3,6 @@ local server = require('nvim-lsp-installer.server')
 local root_dir = server.get_server_root_path('go')
 
 local install_cmd = [=[
-if ! command -v go &> /dev/null;
-then 
-  echo "Please install the Go CLI before installing gopls.";
-  echo "refer to https://golang.org/doc/install";
-  exit 1;
-fi
-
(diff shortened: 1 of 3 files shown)


lua/nvim-lsp-installer/servers/gopls.lua
```diff
@@ -0,0 +1,31 @@
...
+
+local root_dir = server.get_server_root_path('go')
+
+local install_cmd = [=[
+if ! command -v go &> /dev/null;
+then 
+  echo "Please install the Go CLI before installing gopls.";
+  echo "refer to https://golang.org/doc/install";
+  exit 1;
+fi
```
williamboman, line 11: I think you could move this to the `pre_install_check` function instead, something like:

```lua
return server.Server:new {
    ...,
    pre_install_check = function ()
        if vim.fn.executable("go") ~= 1 then
            error("Please install the Go CLI before installing gopls (https://golang.org/doc/install).")
        end
    end
}
```

lua/nvim-lsp-installer/servers/gopls.lua
```diff
@@ -0,0 +1,31 @@
...
+
+GO111MODULE=on go get golang.org/x/tools/gopls@latest;
```
williamboman, line 13: Not super familiar with go - but would it be possible to install the package inside the current working dir?

I'm thinking something like:

```sh
GO111MODULE=on GOPATH="$PWD" go get golang.org/x/tools/gopls@latest
```

The reason for this is to avoid polluting the system and have the installation confined within the LSP server's root_dir.

Then the `cmd` below would need to become something like:

```lua
cmd = { root_dir .. "/bin/gopls" }
```

What do you think?

williamboman (commented):
Hey! Added some thoughts about the installation directory

lua/nvim-lsp-installer/servers/gopls.lua
ecmma: Nice! Much much cleaner! Adding it right away. 
-> added with https://github.com/ecmma/nvim-lsp-installer/commit/5eb4be95b86f3c558eef0b815ec8041fba60f917

lua/nvim-lsp-installer/servers/gopls.lua
ecmma: After some searching, I came up with https://github.com/ecmma/nvim-lsp-installer/commit/747739412f110c701685605e7af12bdb7b689d76. 

However, I noticed that the server doesn't start unless I close and reopen the *.go file, while the normal behaviour is directly start the server, and right now I don't know what causes this (but I don't think it's related to this change.)

lua/nvim-lsp-installer/servers/gopls.lua
```diff
@@ -0,0 +1,29 @@
...
+return server.Server:new {
+  name = "gopls",
+  root_dir = root_dir,
+  pre_install_check = function ()
+    if vim.fn.executable("go") ~= 1 then
+      error("Please install the Go CLI before installing gopls (https://golang.org/doc/install).")
+    end
+  end,
+  install_cmd = install_cmd,
+  default_options = {
```
ecmma, line 26: Also, just a reference for posterity, the server doesn't start unless a `.git` or a `go.mod` is found (as shown [here](https://github.com/neovim/nvim-lspconfig/blob/master/CONFIG.md#gopls)), which prevents the server to be used, for example, in a main.go in your $HOME (unless itself is a go.mod or contains .git) and I think this behaviour can be changed adding 
`root_dir = root_pattern("go.mod", ".git") or dirname`
to the config. I write this here just because I was clueless at first.

lua/nvim-lsp-installer/servers/gopls.lua
```diff
@@ -0,0 +1,29 @@
...
+local install_cmd = [=[
+GO111MODULE=on GOBIN="$PWD" GOPATH="$PWD"  go get golang.org/x/tools/gopls@latest;
+
+if ! command -v ./gopls &> /dev/null; 
+then 
+  echo "something went wrong!"
+else 
+  echo "installation completed"
+fi
+
```
williamboman, line 14: I think this could be simplified to just:

```suggestion
command -v ./gopls &> /dev/null;
```

Reason being that the install scripts are executed with [`set -e`](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/server.lua#L154), meaning it'll exit early if a command exits with a non-0 exit code. `command -v` will exit with code `1` if the specified command does not exist.

As for echoing feedback, this is already taken care of by the installer, [here](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/server.lua#L159) and [here](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/server.lua#L162)

williamboman (commented):
One final thing regarding checking that things were installed correctly, I think it could be made a bit simpler

lua/nvim-lsp-installer/servers/gopls.lua
williamboman: > However, I noticed that the server doesn't start unless I close and reopen the *.go file, while the normal behaviour is directly start the server, and right now I don't know what causes this (but I don't think it's related to this change.)

Hm sounds a bit odd but yeah probably not related to this PR. Could it be that the buffer doesn't connect itself to the LSP server, rather than the server not starting? Maybe something's up with your `on_attach` function 🤔

lua/nvim-lsp-installer/servers/gopls.lua
```diff
@@ -0,0 +1,24 @@
+local server = require('nvim-lsp-installer.server')
+
+local root_dir = server.get_server_root_path('go')
+
+local install_cmd = [=[
+
+GO111MODULE=on GOBIN="$PWD" GOPATH="$PWD"  go get golang.org/x/tools/gopls@latest;
+! command -v ./gopls &> /dev/null;
```
williamboman, line 8:
```suggestion
command -v ./gopls &> /dev/null; 
```

lua/nvim-lsp-installer/servers/gopls.lua
ecmma: https://github.com/ecmma/nvim-lsp-installer/commit/4e461d619c24b52f3fae8797cd0333a1cfa6ac82 nice, thanks. 
Forgot to remove develop-time checks :)

williamboman (approved):
Ty once again @ecmma, great stuff!

lua/nvim-lsp-installer/servers/gopls.lua
ecmma: yeah, thanks! my fault :)

PR #12, opened 2021-05-15
add texlab

ecmma:
Hello, again!

This adds a basic texlab configuration to the plugin. Tested on linux, but should work fine on macOS as well. Technically, there are binaries for Windows too, but it seems virtually unsupported (but it wouldn't be too much of a hassle to add, if needed.)
Also, do you think a check for `tar` is needed?

P.S. Sorry for the logfile in gopls :P

### DIFF
--- a/lua/nvim-lsp-installer/servers/texlab.lua
+++ b/lua/nvim-lsp-installer/servers/texlab.lua
@@ -20,6 +20,13 @@ return server.Server:new {
   name = "texlab", 
   root_dir = root_dir, 
   install_cmd = install_cmd, 
+  pre_install_check = function ()
+    if vim.fn.executable("wget") ~=1 then 
+      error ("Missing wget. Please, refer to https://www.gnu.org/software/wget/ to install it.")
+    elseif vim.fn.executable("pdflatex") ~=1 then
+      error("The program pdflatex wasn't found. Please install a TeX distribution: https://www.latex-project.org/get/#tex-distributions")
+    end
+  end,
   default_options = {
     cmd = {root_dir .. '/texlab'},
   }
--- a/lua/nvim-lsp-installer/servers/texlab.lua
+++ b/lua/nvim-lsp-installer/servers/texlab.lua
@@ -1,0 +1,26 @@
+local server = require('nvim-lsp-installer.server')
+
+local root_dir = server.get_server_root_path('latex')
+
+local install_cmd = [=[
+if [[ $(uname) == Linux ]]; then
+  wget -O texlab.tar.gz https://github.com/latex-lsp/texlab/releases/download/v2.2.2/texlab-x86_64-linux.tar.gz
+elif [[ $(uname) == Darwin ]]; then 
+  wget -O texlab.tar.gz https://github.com/latex-lsp/texlab/releases/download/v2.2.2/texlab-x86_64-macos.tar.gz
+else 
+  >&2 echo "$(uname) not supported."; 
+  exit 1;
+fi
+
+tar xf texlab.tar.gz
+
+]=]
+
+return server.Server:new {
+  name = "texlab", 
+  root_dir = root_dir, 
+  install_cmd = install_cmd, 
+  default_options = {
+    cmd = {root_dir .. '/texlab'},
+  }
+}
--- a/lua/nvim-lsp-installer/servers/texlab.lua
+++ b/lua/nvim-lsp-installer/servers/texlab.lua
@@ -22,7 +22,7 @@ return server.Server:new {
   install_cmd = install_cmd, 
   pre_install_check = function ()
     if vim.fn.executable("wget") ~=1 then 
-      error ("Missing wget. Please, refer to https://www.gnu.org/software/wget/ to install it.")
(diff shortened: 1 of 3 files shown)


williamboman (approved):
LGTM! Do you think it'd make sense to have a `pre_install_check` where we try to detect whether a TeX distribution is installed in PATH, and if it's not, refer to https://www.latex-project.org/get/#tex-distributions in an error message?

williamboman:
> Also, do you think a check for `tar` is needed?

Nah I think we can assume these more common binaries are installed :). I think from the binaries used, `wget` is actually the one that usually causes some issues due to it not shipping in all Linux distros.. or something like that

ecmma:
https://github.com/williamboman/nvim-lsp-installer/pull/12/commits/a0dcadafa2195a4c723fcad2a8b373a99a8069fd adds  a check for boith `wget` and `pdflatex`, which I think is installed with every TeX distribution.

lua/nvim-lsp-installer/servers/texlab.lua
```diff
@@ -0,0 +1,33 @@
...
+
+]=]
+
+return server.Server:new {
+  name = "texlab", 
+  root_dir = root_dir, 
+  install_cmd = install_cmd, 
+  pre_install_check = function ()
+    if vim.fn.executable("wget") ~=1 then 
+      error ("Missing wget. Please, refer to https://www.gnu.org/software/wget/ to install it.")
```
williamboman, line 25:
```suggestion
      error("Missing wget. Please, refer to https://www.gnu.org/software/wget/ to install it.")
```

williamboman (approved):

williamboman:
Ty @ecmma ❤️!

Final question: How feasible do you think it'd be to install a preferred TeX dist, instead of just the lsp? I kinda want this plugin to be as exhaustive as possible so that all you'd need is `:LspInstall` and not depend on some other system installs

ecmma:
I get what you mean. Speaking from a Linux point of view, I think it's better to let the system package manager deal with the dependencies (for example, Linux needs Perl as well, which we should install), and I don't think enumerating each possible manager is viable. 
If we are to implement this (bypassing pacman's), I guess the most "user-respectful" way to do this is ask the user if they want to install it. What do you think?

williamboman:
Yes for sure, opt-in through interactive install scripts. On that note, I've been meaning to try out https://github.com/google/zx for a while, might look into enabling it for the install scripts.

ecmma:
Alright! Seems good to me. Let's wait for zx, then!

williamboman:
I went for it - https://github.com/williamboman/nvim-lsp-installer/pull/13. Huge PR, but wdyt?

Examples:
 - [texlab server definition](https://github.com/williamboman/nvim-lsp-installer/blob/6901fb3d7e02e74d7888c01861c76e33078ede39/lua/nvim-lsp-installer/servers/texlab/init.lua)
 - [texlab zx install script](https://github.com/williamboman/nvim-lsp-installer/blob/6901fb3d7e02e74d7888c01861c76e33078ede39/lua/nvim-lsp-installer/servers/texlab/install.mjs)

ecmma:
That was fast! Nice! It's cleaner than a string containing a command, that's for sure. I'm trying my hand at `zx`, but we need root premissions to write /usr/local/texlive (and install perl, I guess) and to install MacTeX on macOS: 
>  Indeed, we recommend installing it as a normal user, except perhaps on MacOSX, where it's conventional to install as administrator. (Some information about shared installations.) As always, you need to have permission to write into the destination directory, but TeX Live itself does not care if you are root or not.

(from [here](https://www.tug.org/texlive/quickinstall.html))
Any idea about that?

On a side-note, have you tried the [`question` function](https://github.com/google/zx#question) provided by zx and in particular the {choices: } option? I can't seem to get it to work as it accepts any choice, even in their example: 

``` js
#!/usr/bin/env zx

let token = await question('Choose env variable: ', {
  choices: Object.keys(process.env)
})

console.log(token)
```

``` bash 
$ ./test.mjs                                                                                                                                                                                                                         
Choose env variable: NO_ENV_VAR
NO_ENV_VAR
$ [ -z "NO_ENV_VAR" ]                   [1]                                                                                                                                                                
```

but maybe I'm expecting something that it simply doesn't do :P

williamboman:
> Any idea about that?

I guess what they're saying is that in order to install to default install locations (under `/usr/local`) you generally need root permissions. But the software itself doesn't care where you install it, as long as it's in your `$PATH`.

I however think we should avoid installing things on the system outside of the allocated `vim.fn.stdpath("data") .. "/lsp_servers"` directory. So in the case of installing a TeX dist, we should put it in the root directory of the texlab server (e.g. `~/.local/share/nvim/lsp_servers/latex`). This probably means we'd also have to add this install directory to the `PATH` before starting the LSP.

> On a side-note, have you tried the [`question` function](https://github.com/google/zx#question) provided by zx and in particular the {choices: } option? I can't seem to get it to work as it accepts any choice, even in their example:

Huh I noticed that as well when I was trying it out. I thought I had screwed something up and didn't think much more of it.

PR #13, opened 2021-05-16
add support for zx install scripts

williamboman:
# Why?

1. Nicer scripting language than the shell. Suitable for more complicated installs where lots of plumbing needs to be done.
2. Allow for more sophisticated install scripts, for example interactive ones with different branches controlled by user input.
3. Potentially broader platform support out of the box.
4. Make it easier to write test suites for these install scripts and hook them up to CI and whatnot.
5. JS/TS has a much better ecosystem for static analysis.


lua/nvim-lsp-installer/fs.lua
```diff
@@ -0,0 +1,42 @@
+local uv = vim.loop
```
abderrahmane-tj, line 1: are you sure you're not re-inventing the wheel here?

lua/nvim-lsp-installer/path.lua
```diff
@@ -0,0 +1,38 @@
+local uv = vim.loop
+
+local sep = (function()
```
abderrahmane-tj, line 3: are you sure no helper already exists? otherwise, couldn't you use `os.platform` and for win32 use `\` otherwise `/` ?

lua/nvim-lsp-installer/servers/clangd/install.mjs
```diff
@@ -0,0 +1,20 @@
+const VERSION = "11.0.0";
```
abderrahmane-tj, line 1: not sure actually needed, but no try catching and nice errors here? or that's too overkill?

lua/nvim-lsp-installer/fs.lua
williamboman: `vim.loop` (libuv) is pretty much the only Lua API for filesystem operations, that I am aware of. And it's very low level, hence this module. Plugins like [plenary.nvim](https://github.com/nvim-lua/plenary.nvim) exists to make it easier to work with.

lua/nvim-lsp-installer/path.lua
williamboman: Pretty sure. This was copied from plenary.nvim actually.

>  couldn't you use os.platform and for win32 use \ otherwise / ?

Isn't that pretty much exactly what it's doing right now though?

lua/nvim-lsp-installer/servers/clangd/install.mjs
williamboman: zx runs in "`set -e`" mode [by default](https://github.com/google/zx#prefix), so if a command fails its promise will just reject and abort the process. The Lua code that starts the script will then handle [handle errors](https://github.com/williamboman/nvim-lsp-installer/blob/zx/lua/nvim-lsp-installer/server.lua#L153)

abderrahmane-tj:
Well, you're naming all os types under the sun though 

On May 16, 2021 11:26:31 PM GMT+02:00, William Boman ***@***.***> wrote:
***@***.*** commented on this pull request.
>
>
>
>> @@ -0,0 +1,38 @@
>+local uv = vim.loop
>+
>+local sep = (function()
>
>Pretty sure. This was copied from plenary.nvim actually.
>
>>  couldn't you use os.platform and for win32 use \ otherwise / ?
>
>Isn't that pretty much exactly what it's doing right now though?
>
>-- 
>You are receiving this because you commented.
>Reply to this email directly or view it on GitHub:
>https://github.com/williamboman/nvim-lsp-installer/pull/13#discussion_r633141557

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #24, opened 2021-06-11
add get_server({server_name}) to public Lua API

### LINKED ISSUE #22: [Feature Request] Expose get_server function
I'm trying to do something like this:
```lua
local lsp_installer = require('nvim-lsp-installer')
local _, server = lsp_installer.get_server("rust-analyzer")

if not server.is_installed() then
    server.install()
end
```

I want to do this so that I only install the language server if a specific plugin is loaded (in this case `rust-tools.nvim`).

But I'm unable to because the `get_server` function isn't exposed. Would it be possible to expose it?

williamboman:
Closes #22.

### DIFF
--- a/doc/nvim-lsp-installer.txt
+++ b/doc/nvim-lsp-installer.txt
@@ -88,6 +88,23 @@ get_uninstalled_servers()
         Return: ~
             |lsp_installer.Server|[] A list of servers that are not installed.
 
+                                                 *lsp_installer.get_server()*
+get_server({server_name})
+        Parameters: ~
+            {server_name} (string) The server to install.
+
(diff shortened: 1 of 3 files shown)


williamboman:
@Akeboshiwind wdyt?

Akeboshiwind:
Nice, that looks good to me.

It's much nicer to have this as part of the API rather than having to reach
for the internals :D

williamboman:
> It's much nicer to have this as part of the API rather than having to reach for the internals :D

Yeah for sure!

Akeboshiwind:
Thanks for the quick turn around :D

PR #26, opened 2021-06-19
texlab: upgrade to v3.2.0

williamboman:


williamboman:
@ecmma Would you happen to have the time to test this out? I can't be bothered re-installing a distribution all over again 😅

ecmma:
Just checked, changing `install_cmd` to `installer` in texlab/init.lua  all works well! Thanks for the good job :)

PR #60, opened 2021-08-22
export proper location of haskell language servers

mystilleef:
`${__dirname}` was exporting the location of the `Packer` installation folder, instead of the location of the Haskell language servers and `hls` script. This change fixes the issue for me.

### DIFF
--- a/lua/nvim-lsp-installer/servers/hls/install.mjs
+++ b/lua/nvim-lsp-installer/servers/hls/install.mjs
@@ -22,7 +22,9 @@ await $`rm hls.tar.gz`;
 await $`chmod +x haskell*`;
 
 const scriptContent = `#!/usr/bin/env bash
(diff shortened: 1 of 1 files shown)


lua/nvim-lsp-installer/servers/hls/install.mjs
```diff
@@ -22,7 +22,9 @@ await $`rm hls.tar.gz`;
 await $`chmod +x haskell*`;
 
 const scriptContent = `#!/usr/bin/env bash
-PATH="$PATH:${__dirname}" "${__dirname}/haskell-language-server-wrapper" --lsp`;
```
williamboman, line 25: Ah! I think this was originally supposed to be `process.cwd()` instead of the `__dirname` of the script itself

williamboman (approved):
thanks!

PR #73, opened 2021-09-02
add rescriptls

aspeddro:


lua/nvim-lsp-installer/servers/rescriptls/init.lua
```diff
@@ -0,0 +1,21 @@
...
+return server.Server:new {
+    name = "rescriptls",
+    root_dir = root_dir,
+    installer = shell.polyshell [[
+        curl -s https://api.github.com/repos/rescript-lang/rescript-vscode/releases/latest \
+              | grep "browser_download_url.*vsix" \
+              | cut -d : -f 2,3 \
+              | tr -d '"' \
+              | wget -i - -O vscode-rescript.vsix;
+        unzip -q -o vscode-rescript.vsix;
```
williamboman, line 16: Overall this LGTM! One thing though - this does not read like cmd.exe compatible to me, making it not compatible for the polyshell installer.

I think we should change this to
```suggestion
    installer = installers.when {
        unix = shell.bash [[
           curl -s https://api.github.com/repos/rescript-lang/rescript-vscode/releases/latest \
                  | grep "browser_download_url.*vsix" \
                  | cut -d : -f 2,3 \
                  | tr -d '"' \
                  | wget -i - -O vscode-rescript.vsix;
            unzip -q -o vscode-rescript.vsix;
        }
```

which effectively makes this available to UNIX systems only

lua/nvim-lsp-installer/servers/rescriptls/init.lua
```diff
@@ -0,0 +1,23 @@
...
+local shell = require("nvim-lsp-installer.installers.shell")
+
+local root_dir = server.get_server_root_path('rescriptls')
+
+return server.Server:new {
+    name = "rescriptls",
+    root_dir = root_dir,
+    installer = installers.when {
+        unix = shell.bash [[
+           curl -s https://api.github.com/repos/rescript-lang/rescript-vscode/releases/latest \
```
williamboman, line 12: Let's add `-f` so we fail early and don't pipe potential error responses

```suggestion
           curl -fs https://api.github.com/repos/rescript-lang/rescript-vscode/releases/latest \
```

williamboman (approved):
Thanks!

PR #81, opened 2021-09-08
add r lang server

aspeddro:
Add R language server.

This install the library `languageserver` globally. Specifically in `.libPaths()`


williamboman (commented):
Cool!! Is there any particular reason why it's installed globally? Or is it simply the only way to install it? I'd have hoped that this plugin would be able to never install, or in any way modify the system, outside of the dedicated `lsp_servers` directory inside the `data` stdpath. Also, installing it globally introduces some awkwardness with some of the other functionalities:

1. determining whether it's still installed or not
2. uninstalling an LSP server

As well as some other things I have in mind for the future, such as detailed stats about each LSP server, like:
- disk space used
- managed versioning to allow detecting whether an upgrade is available
- multiple instances of the same server, targeting different use cases (e.g. different versions)

I think we might need to consider adding some kind of other construct for expressing a LSP server that will be installed globally, to conform to at least the first two points above. I'll give it some more thought!

lua/nvim-lsp-installer/servers/r_language_server/init.lua
```diff
@@ -0,0 +1,16 @@
+local server = require 'nvim-lsp-installer.server'
+local shell = require "nvim-lsp-installer.installers.shell"
+local path = require "nvim-lsp-installer.path"
+
+local root_dir = server.get_server_root_path("r_language_server")
+
+return server.Server:new {
+    name = "r_language_server",
+    root_dir = root_dir,
+    installer = shell.polyshell([[R -e 'install.packages("languageserver", repos="http://cran.us.r-project.org")']]),
```
williamboman, line 10: This brings some extra custom code but! We can avoid going through the shell as the "mediator" by spawning a process directly instead, like:

```lua
local process = require("nvim-lsp-installer.process")
-- ...
    installer = function (server, callback, context)
        process.spawn("R", {
          args = { "-e", "'install.packages("languageserver", repos="http://cran.us.r-project.org")' },
          stdio_sink = context.stdio_sink
        }, callback)
    end
```

This will go through Lua APIs exclusively (using the `vim.loop (libuv)` API)

aspeddro:
@williamboman now install in `root_dir`

williamboman (approved):
I tried this out on my Mac and the installation went through perfectly (although took a surprisingly long time). However, when I try it out in a random R project that I found on github the LSP server crashes - the logs mentioned something about `langserver` not being available (I'm on another machine atm so can't get the exact logs).

But if this works for you I'm happy to merge (perhaps after #85 goes through though)!

williamboman:
@aspeddro I pushed some commits to make this compatible with some changes I've merged in `main` branch.

I'm still getting this error though:

```
[ ERROR ] 2021-09-17T20:40:44+0200 ] /opt/nvim/share/nvim/runtime/lua/vim/lsp/rpc.lua:462 ]	"rpc"	"R"	"stderr"	'Error in library(languageserver, lib.loc = "/Users/williamboman/.local/share/nvim/lsp_servers/r_language_server") : \n  there is no package called ‘languageserver’\nExecution halted\n'
```

I have no experience with R whatsoever so I might be doing something weird - could you verify that this works for you?

aspeddro:
@williamboman it's working. `languageserver` has several dependencies with bindings to `C` and `C++` libraries. I think it was not installed due to some dependency and compilation failed.

You can try installing from the console to get the log. 
```r
install.packages("languageserver", repos="https://cloud.r-project.org", lib="/Users/williamboman/.local/share/nvim/lsp_servers/r_language_server")
```
If at the end it shows `DONE (languageserver)` then everything went right.

image

I'm going to change the install process to clone the repo and build the package. The version on CRAN is out of date.

aspeddro:
@williamboman, I did some testing and installing `languageserver` in `root_dir` is not fully isolated. R does not install `languageserver` dependencies on `root_dir` if it is already installed globally.

It is possible to isolate only `languageserver` not its dependencies. Dependencies are installed globally. 

I think not a good case for this plugin.

williamboman:
I did some more investigation and gave the following a try, but it seems like it doesn't really work. I'm struggling to test this actually because there is some dependency that won't compile on my M1 Mac. Will give it one final try, but otherwise yeah it'd not be a good fit for this plugin unfortunately :(

```diff
diff --git a/lua/nvim-lsp-installer/servers/r_language_server/init.lua b/lua/nvim-lsp-installer/servers/r_language_server/init.lua
index 2cb0fee..6a1e9f1 100644
--- a/lua/nvim-lsp-installer/servers/r_language_server/init.lua
+++ b/lua/nvim-lsp-installer/servers/r_language_server/init.lua
@@ -9,20 +9,21 @@ return function(name, root_dir)
             process.spawn("R", {
                 args = {
                     "-e",
-                    ('if (!require("remotes")) { install.packages("remotes", repos="https://cloud.r-project.org") };') ..
-                    ('remotes::install_github("REditorSupport/languageserver", upgrade="default", force=TRUE, lib=%q)'):format(
-                        root_dir
-                    ),
+                    ('install.packages("languageserver", repos="https://cloud.r-project.org", lib=%q)'):format(root_dir),
+                },
+                env = process.graft_env {
+                    R_LIBS = root_dir,
+                    R_LIBS_USER = root_dir,
+                    R_LIBS_SITE = root_dir,
                 },
                 stdio_sink = context.stdio_sink,
             }, callback)
         end,
         default_options = {
-            cmd = {
-                "R",
-                "--slave",
-                "-e",
-                ("library(languageserver, lib.loc = %q); languageserver::run()"):format(root_dir),
+            cmd_env = {
+                R_LIBS = root_dir,
+                R_LIBS_USER = root_dir,
+                R_LIBS_SITE = root_dir,
             },
         },
     }
```

williamboman:
@aspeddro :wave: did the latest commit make this work for you? Or does it still partially install globally as before?

aspeddro:
@williamboman using environment variable solved the problem to isolate dependencies entirely, but it changes the path of the `.libPaths()` libraries, so `languageserver` doesn't index installed packages globally. The user would have to install all libraries in `root_dir`.

I'll close. Thank you for your help.

PR #85, opened 2021-09-09
rewrite some installers for broader cross-platform support

williamboman:
Also adds full support for Windows for all installers and other improvements.


williamboman:
Hey @mystilleef 👋! Would you mind trying out this branch for Haskell specifically? It's pretty much the only language server that I have been unable to test, apart from just the installation itself.

The change I made in this PR was that we run the provided executable directly, instead of going through the custom script file that gets created. Instead, we provide the modified `PATH` in the `cmd_env` field.

mystilleef:
I did a quick test. It seems to work. Although, I've just resorted to managing the LSPs myself since for some I want to track the `main` branch myself. For `hls` in particular, it has to compiled with the exact `GHC` version for it to work properly. So it makes sense for me to manage it myself.

williamboman:
> I did a quick test. It seems to work.

Great, thanks!

> For hls in particular, it has to compiled with the exact GHC version for it to work properly. So it makes sense for me to manage it myself.

Gotcha! You could probably do this by writing a custom server for `hls` and registering it with nvim-lsp-installer, if you want to manage it from within neovim!

PR #94, opened 2021-09-17
add jdtls

williamboman:


williamboman:
@ChristianChiarulli 👋 I heard it through the grapevine that you thought jdtls was missing in this plugin :). I did quick tests and it at least comes alive, but I'm not sure if something needs to be fine tuned. I'd be really happy if you could try this branch out as I don't know anyone who uses neovim for Java dev who could verify this for me.

To get going this is the minimum you need to actually start servers installed via nvim-lsp-installer:

```lua
require("nvim-lsp-installer").on_server_ready(function (server)
  server:setup {} -- proxy function for lspconfig's setup() function, takes same args
  vim.cmd [[ do User LspAttachBuffers ]]
end)
```

PR #97, opened 2021-09-21
installers/npm: ensure local npm root dir before installing

### LINKED ISSUE #96: LspInstall installs nothing and cannot easily see installation logs
```
:LspInstall pyright
```

After executing this, `:LspInstallInfo` says that the `pyright` server is installed successfully (installed today).

However, lsp servers can't start:

```
cmd ["$HOME/.local/share/nvim/lsp_servers/python/node_modules/.bin/pyright-langserver"] is not executable.
```

In the directory `$HOME/.local/share/nvim/lsp_servers/` folders like `python`, `tsserver`, etc. are created but they are **just empty directories**.

FYI, I am using macOS and my neovim version is `NVIM v0.6.0-dev+1556-gc642f3ac2                                                                                                  `.

Plus, I cannot find any relevant log files; I wanted to show **the output of installer scripts** but apparently the GUI window doesn't have any feature. How can I debug this?

williamboman:
Resolves #96.

### DIFF
--- a/lua/nvim-lsp-installer/installers/npm.lua
+++ b/lua/nvim-lsp-installer/installers/npm.lua
@@ -28,15 +28,13 @@ function M.packages(packages)
             cwd = server.root_dir,
             stdio_sink = context.stdio_sink,
         }
-        if
-            not (
-                fs.dir_exists(path.concat { server.root_dir, "node_modules" })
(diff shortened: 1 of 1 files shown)


wookayin:
Confirmed this works as expected and fixes #96 in the scenario where any of the ancestor directories had `package.json`.

PR #98, opened 2021-09-22
feat: use <esc> to quickly dismiss the info panel

kylo252:


lua/nvim-lsp-installer/ui/display.lua
```diff
@@ -137,6 +137,9 @@ function M.new_view_only_win(name)
             vim.cmd(("autocmd %s <buffer> call v:lua.lsp_install_redraw(%d)"):format(redraw_event, win))
         end
 
+        vim.api.nvim_buf_set_keymap(buf, "n", "<esc>", "<cmd>bd<CR>", { noremap = true })
+        vim.lsp.util.close_preview_autocmd({ "BufHidden", "BufLeave" }, win)
```
williamboman, line 141: What does this do?

williamboman: Aah, nvm ok. Hmm not sure about this one. I think there's a use case for keeping the window open while navigating to other windows, to allow following installation progress

williamboman: I'm thinking that this behavior is somewhat contradictory to at least what I'd considered normal behavior (I'm thinking for example quickfix, loclist, `packer.nvim`). None of these takes control of window management as such. I think this kind of behavior would feel more natural in a floating window setting

kylo252: It can also be similar to how Packer does it (it's actually a good idea to check the state beforehand) https://github.com/wbthomason/packer.nvim/blob/0a2d8cbaa2045bdf3797af7a5abb2d42d0edecb0/lua/packer/display.lua#L847

williamboman: Not sure I'm following, do you mean cancelling all server installations if a user closes the window?

TeoDev1611: That's sound cool can be cancel the installations :)

kylo252: > Not sure I'm following, do you mean cancelling all server installations if a user closes the window?

It would prevent the panel from closing until it's "okay" to do so. This might be necessary until the `status_ui` is decoupled from the installation process.

> That's sound cool can be cancel the installations :)

I don't think that's a good idea actually, it's really hard to cleanup such cases and the installer barely takes that much time anyway.

williamboman: > It would prevent the panel from closing until it's "okay" to do so. This might be necessary until the `status_ui` is decoupled from the installation process.

Ah gotcha! I think that could make sense. I am however a bit hesitant to anything that has to do with automatic, "smart", window management. I think neovim users are not the right target group for that :D. I feel like this kinds of aims to make the window behave as a popup/floating window, in which case I think we should go all the way and do actual floating windows, and not misuse the window system?

Is there a specific interop problem with LunarVim you're trying to solve here or is it mostly preferences?

> I don't think that's a good idea actually, it's really hard to cleanup such cases and the installer barely takes that much time anyway.

I agree, I'd find that behavior a bit weird. I think of these all as background operations that have a UI that provides a cursory overview.

kylo252: It's definitely a usability/accessibility setting for me. 

Look what happens if I try to open it with a more compact terminal width, and how it can get out of hand if you happen to have nvim-tree opened: https://asciinema.org/a/6cjmyrhLVMdu1eTTG7CHvkFWd

It's fundamentally a vim problem, commonly referred to "[oil vs vinegar](https://github.com/tpope/vim-vinegar)". Even on a full-screen terminal, the `LspInstallInfo` panel is taking half the real-estate.

williamboman: Yeah I see.. Let's go with this for now, but try to implement a proper floating window (which would have the similar behavior as this, but.. floating)

PR #99, opened 2021-09-22
feat: add a logger based on plenary.log

kylo252:
## Description

Add a logger based on Plenary's log implementation.

I basically hijacked the notify function, so it's trivial to swap out the logger for a more sophisticated one, should you wish so.

--- 

_Note: It's up to you to decide on how to set the default log level._

### DIFF
--- a/lua/nvim-lsp-installer/log.lua
+++ b/lua/nvim-lsp-installer/log.lua
@@ -1,8 +1,4 @@
---
--- The following is extracted and modified from plenary.log by TJ Devries.
-
--- User configuration section
-local default_config = {
+local config = {
     -- Name of the plugin. Prepended to log messages
     name = "lsp-installer",
 
@@ -15,9 +11,6 @@ local default_config = {
 
     -- Should write to a file
     use_file = true,
-
-    -- Any messages above this level will be logged.
-    level = "warn",
 
     -- Level configuration
     modes = {
@@ -33,27 +26,17 @@ local default_config = {
     float_precision = 0.01,
 }
 
--- {{{ NO NEED TO CHANGE
 local log = {}
+
+log.levels = vim.deepcopy(vim.log.levels)
+
+-- Default log level is warn.
+local current_log_level = log.levels.WARN
 
 local unpack = unpack or table.unpack
 
-log.new = function(config, standalone)
-    config = vim.tbl_deep_extend("force", default_config, config)
-
+do
     local outfile = string.format("%s/%s.log", vim.api.nvim_call_function("stdpath", { "cache" }), config.name)
-
-    local obj
-    if standalone then
-        obj = log
-    else
-        obj = config
-    end
-
-    local levels = {}
-    for i, v in ipairs(config.modes) do
-        levels[v.name] = i
-    end
 
     local round = function(x, increment)
         increment = increment or 1
@@ -80,8 +63,8 @@ log.new = function(config, standalone)
     end
 
     local log_at_level = function(level, level_config, message_maker, ...)
-        -- Return early if we're below the config.level
-        if level < levels[config.level] then
+        -- Return early if we're below the current_log_level
+        if level < current_log_level then
             return
         end
         local nameupper = level_config.name:upper()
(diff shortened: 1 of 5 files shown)


kylo252:
Here's a demo of what it looks like in action:

jZOhTGEC89

williamboman:
Cool!! This is something I had in mind. One thing that kept me putting this of is the log-to-storage part. I glanced through the code so I think I have the answer, but is there any log rotation/pruning in place? Bit worried that these log files are unbounded in size and will just keep growing forever, which is not very cash money. The amount of logging that is in place today is no cause for concern though, but I was planning on adding a bunch more (like performance metrics and whatnot).

lua/nvim-lsp-installer/log.lua
```diff
@@ -1,18 +1,175 @@
-local M = {}
```
williamboman, line 1: Do we need to attribute this code to the original author(s) here? It looks like it's under the MIT license

lua/nvim-lsp-installer/notify.lua
```diff
@@ -8,6 +8,12 @@ return function(msg, level)
...
         })
     else
-        vim.notify(("[%s] %s"):format(TITLE, msg), level)
+        local vim_log_levels = vim.tbl_keys(vim.log.levels)
+        local log_levels = vim.tbl_map(function(lv)
+            return lv:lower()
+        end, vim_log_levels)
+
+        local Log = require "nvim-lsp-installer.log"
+        Log[log_levels[level]](msg)
```
williamboman, line 17: I'm not sure this needs changing. In my mind, notifications to the user (this module) are very different than system logging (my idea of `log.lua`).

[`vim.notify()` already takes care of setting highlight group based on log level](https://github.com/neovim/neovim/blob/master/src/nvim/lua/vim.lua#L421-L430), so I think going through the log is redundant here, no?

lua/nvim-lsp-installer/log.lua
kylo252: It's totally fine to remove it 👍

lua/nvim-lsp-installer/notify.lua
kylo252: Keep in mind, you can create another instance of the logger that doesn't write to a file. 

This API is a little bit more flexible, unless you're very set on the use-case of `notify`, with the main benefit being that it allows three different modes: { 'sync', 'async', false }.

williamboman: I don't have any strong opinions (or well, I might but they're loosely held) on this. I fail to see the reason to substitute usage of the native `vim.notify` for this logger, for messages intended to be displayed to the user - it just adds overhead with no benefits unless I am missing something. As for "system" logs stored to file, I definitely see a need for a logger implementation like this, but again I'd like to keep these separate from one another.

What I'm picturing is 
1. keeping the `notify` module as-is (i.e., keep going through `vim.notify` directly for message intended for users)
2. change the `log` module to only act as a system logger by default (no `echo`, only `io.write`)

Am I missing something here or does that make sense? I'd be happy to help contribute with above changes and at the meantime add some more logging + a new `vim.g` option to control the system log level + docs.

kylo252: Sounds good, it's an easy swap like I said. 

The only thing really that's left in this case is deciding on some kind of method to set the log-level. You can always use a global variable with `vim.g`, but I also think it might be worth it to consider using an `init()` function soon.

williamboman: I think a global variable, that can be changed and have direct impact throughout a neovim session, is pretty nice for this particular setting, no? To temporarily increase logging in order to debug something for example. No need to change your own Lua code. I was never a fan of plugins having `setup()` or `init()` functions tbh, but is that where the direction is generally going?

kylo252: > edit: Ah the logger impl doesn't allow changing log level after instantiation

Yeah that's the problem, so it's not a toggle. That's why I figured it would make sense to have in a setup function. I'll try something else and see if it works.

williamboman: I noticed earlier that it seems like the lua code executes before `plugin/nvim-lsp-installer.vim` (where defaults are defined). That caught me a bit off guard tbh.

kylo252: I think it should be fine now. I tested with with changing the log-level at runtime and it worked correctly for me 😃 

This is now a hybrid between the plenary logger and the one used by the LSP module in core.

williamboman:
@kylo252 I opened a PR to this branch at https://github.com/kylo252/nvim-lsp-installer/pull/1. Apart from those suggestions I think this is ready to merge!

williamboman (approved):

PR #100, opened 2021-09-24
feat(ui): add a popup option

kylo252:
This still needs work but I think it's nearly there 😄 

You might be able to handle some of the quirks better, since you're more familiar with the state handler.


kylo252:
Here's a test with a small terminal space from the beginning

Gi0siaCEJA

I still need to figure out where those weird lines at the top are coming from.

williamboman:
> I still need to figure out where those weird lines at the top are coming from.

This is probably the indent blankspace plugin. You can disable it for certain buftypes and filetypes, like so https://github.com/williamboman/nvim-config/blob/main/lua/wb/plugins.lua#L180-L181. That reminds we should probably set a dedicated filetype for the UI buffer..

williamboman (commented):
Awesome! Let's make this default behavior and remove the old window method entirely. With ephemeral floating windows we can safely couple the buffer with a single window, where we can also safely assume that if the buffer is valid, the window should be too (thanks to `bufhidden=wipe`). With normal windows we can never really assume this as the buffer could be loaded into many (hence why we used to select the first window the buffer is attached to in the render function). Could you also remove the log code from this PR 🙏?

I'll take a look at the state stuff when I get to a computer 🤓

lua/nvim-lsp-installer/ui/display.lua
```diff
@@ -99,42 +99,72 @@ end
 
 function M.new_view_only_win(name)
     local namespace = vim.api.nvim_create_namespace(("lsp_installer_%s"):format(name))
-    local buf, renderer, mutate_state, get_state, unsubscribe
+    local buf_handle, renderer, mutate_state, get_state, unsubscribe, win_id
```
williamboman, line 102: Nitpick, but maybe `bufnr` and `winnr` for consistency?

lua/nvim-lsp-installer/ui/display.lua
```diff
@@ -143,38 +173,40 @@ function M.new_view_only_win(name)
...
         local output = render_node(context, view)
         local lines, virt_texts, highlights = output.lines, output.virt_texts, output.highlights
 
         vim.api.nvim_buf_clear_namespace(0, namespace, 0, -1)
-        vim.api.nvim_buf_set_option(buf, "modifiable", true)
-        vim.api.nvim_buf_set_lines(buf, 0, -1, true, lines)
-        vim.api.nvim_buf_set_option(buf, "modifiable", false)
+        vim.api.nvim_buf_set_option(buf_handle, "modifiable", true)
+        -- vim.api.nvim_buf_set_lines(buf_handle, 0, -1, true, lines)
+        vim.api.nvim_buf_set_lines(buf_handle, 0, -1, false, lines)
```
williamboman, line 197: Huh I wonder why I had put the `strict_indexing` arg to `true`, let's change to `false` :+1:

lua/nvim-lsp-installer/ui/display.lua
```diff
@@ -99,42 +99,72 @@ end
...
+            swapfile = false,
+            textwidth = 0,
+            buftype = "nofile",
+            bufhidden = "wipe",
+            buflisted = false,
+            filetype = "lsp-installer",
+        }
+
+        local win_opts = {
+            number = false,
```
williamboman, line 146: I believe we do need to set `norelativeumber` too, they're not mutually exclusive

lua/nvim-lsp-installer/ui/display.lua
```diff
@@ -99,42 +99,72 @@ end
...
+            local popup_layout = {
+                relative = "editor",
+                height = math.floor(win_height * 0.9),
+                width = math.floor(win_width * 0.8),
+                style = "minimal",
+                border = "rounded",
+            }
+            popup_layout.row = math.floor((win_height - popup_layout.height) / 2)
+            popup_layout.col = math.floor((win_width - popup_layout.width) / 2)
+            buf_handle = vim.api.nvim_create_buf(false, true)
```
williamboman, line 122: I feel like this should be the other way around, either

```suggestion
            buf_handle = vim.api.nvim_create_buf(true, false)
```

or 

```suggestion
            buf_handle = vim.api.nvim_create_buf(true, true)
```

There's no reason for the buffer to be listed, is there?

lua/nvim-lsp-installer/ui/display.lua
kylo252: Funny you should ask, I was actually going for `bufnr` but I thought you specifically avoided them. I had to look it up because I could've sworn it was always `bufnr` in every instance I saw, but no! The official docs actually refers to them as a `handle` rather than a number. So it's a toss up for me, I'll let you decide.

kylo252:
The title was wrong because I wanted to base it on the logger branch for easier debugging, so you can also ignore the logger commits.

kylo252:
> This is probably the indent blankspace plugin. You can disable it for certain buftypes and filetypes, like so [https://github.com/williamboman/nvim-config/blob/main/lua/wb/plugins.lua#L180-L181](https://github.com/williamboman/nvim-config/blob/main/lua/wb/plugins.lua?rgh-link-date=2021-09-24T21%3A02%3A38Z#L180-L181). ~That reminds we should probably set a dedicated filetype for the UI buffer..~ this is already being set apparently

I made the options into a list just so that I can quickly toggle them on and off and I still couldn't figure out what's up with it. I need to see if it's something with the redraw hook.

> Let's make this default behavior and remove the old window method entirely.

That will actually simplify some of the code and probably protect against some potent bugs, since I didn't really heavily test using a split afterwards.

> With ephemeral floating windows we can safely couple the buffer with a single window, where we can also safely assume that if the buffer is valid

If you're set on that then maybe we can also make a meta-table for the popup instead of relying on global state. See this minimal implementation:  https://github.com/LunarVim/LunarVim/blob/9b76a13512b407f23eb9fc16d0210daccccf4802/lua/interface/popup.lua
or something with more bells and whistles: https://github.com/nvim-lua/popup.nvim/blob/f91d80973f80025d4ed00380f2e06c669dfda49d/lua/popup/init.lua

>  Could you also remove the log code from this PR 🙏?

As I said it's for debugging purposes, but sure, I can use them without committing the files 👍

lua/nvim-lsp-installer/ui/display.lua
kylo252: I wasn't sure what this does, that's why I left it commented out so I can remember it 😅

lua/nvim-lsp-installer/ui/display.lua
kylo252: TIL!

lua/nvim-lsp-installer/ui/display.lua
kylo252: Oh man, I do hate having these "magic" values sometimes, bu I've only seen it used like this so I never questioned it.

The way I have it now `vim.api.nvim_create_buf(false, true)` creates a *nobuflisted* and *scratch* buffer, isn't that correct?


```doc
nvim_create_buf({listed}, {scratch})

                Creates a new, empty, unnamed buffer.

                Parameters: ~
                    {listed}   Sets 'buflisted'
                    {scratch}  Creates a "throwaway" |scratch-buffer| for
                               temporary work (always 'nomodified'). Also sets
                               'nomodeline' on the buffer.
```

williamboman: Ahhh you're right.. got me booleans mixed up

lua/nvim-lsp-installer/ui/display.lua
williamboman: I believe it makes it so it errors if you try to set a line that doesn't yet exist in the buffer?

williamboman:
> If you're set on that then maybe we can also make a meta-table for the popup instead of relying on global state. See this minimal implementation: https://github.com/LunarVim/LunarVim/blob/9b76a13512b407f23eb9fc16d0210daccccf4802/lua/interface/popup.lua

Yeah I think this could be nice. However since there's only one UI window currently, maybe let's consider it in the future if there should be a need for more variants of UI windows? I avoided metatable initially because they're so awkward to write from memory. I'm not sure it'll particularly help with state management though? The current state is kept in a closure rather than in say a metatable instance, which to me is just semantics.

> That will actually simplify some of the code and probably protect against some potent bugs, since I didn't really heavily test using a split afterwards.

Yep exactly! Let's do it :+1:

lua/nvim-lsp-installer/ui/display.lua
williamboman: Hehe I remember using `bufnr`, `winnr` at first but later removed the `nr` because I ran into some quirks that suggested there's not a perfect 1:1 match between "Vim" and "Nvim Lua" buffer APIs, but I probably just did something wrong and came to a wrong conclusion, I'm still a bit new to these APIs.

I say if the "handle" terminology is the neovim standard, let's go with `buf_handle`, `win_handle` - otherwise `bufnr`, `winnr`. Again, nitpicking 😅

kylo252: If we're going by the general usage then it's `bufnr` and `win_id`, at least from what I've seen in core and Telescope.

kylo252:
I thought I could do a quick clean-up just to see if there's anything broken. We still have a few things to fix/test.

williamboman:
👍

williamboman:
@kylo252 FYI I created the following PRs to fix some bugs I ran into:
- https://github.com/kylo252/nvim-lsp-installer/pull/3
- https://github.com/kylo252/nvim-lsp-installer/pull/2

kylo252:
So I had to stop at 832e37c because anything after this commit is causing a segfault when pressing `<c-w>`

williamboman:
Hmm I think I might've forgot to push some things actually (probably won't fix segfault, but still).. What system are you on?

williamboman:
From my end I think this is ready to go (apart from the segfault issue of course)

kylo252:
I just tried it on two different machines and both of them are crashing if you attempt to press `<C-w>`. That's bad because it's probably the first button most users reach for when dealing with splits/windows 😢

williamboman:
Haha yes not very good. What OSes and `nvim --version` did you try? These work fine for me:

```
- Windows 10, nvim v0.6.0-dev+112
- macOS Monterey Beta, nvim v0.6.0-dev+344
```

Does increasing `verbose` level help pinpoint the issue maybe?

lua/nvim-lsp-installer/ui/display.lua
```diff
@@ -200,30 +240,34 @@ function M.new_view_only_win(name)
...
                 return
             end
             unsubscribe(false)
-            local opened_win = open(opts)
+            local opened_win_id = open(opts)
             draw(renderer(get_state()))
-            redraw_by_winnr[opened_win] = function()
+            redraw_by_win_id[opened_win_id] = function()
                 draw(renderer(get_state()))
+                vim.api.nvim_win_set_config(opened_win_id, create_popup_window_opts())
```
kylo252, line 260: Here's the redraw logic. We should try to see if we can simplify it a bit as to not have a chance of an invalid `win_id`?

cc @Tastyep

williamboman: Maybe checking `nvim_win_is_valid()`?

kylo252: @williamboman, we should probably introduce a lot of debug prints in this file. It would help with this and in the future.

lua/nvim-lsp-installer/ui/display.lua
```diff
@@ -175,18 +185,12 @@ function M.new_view_only_win(name)
...
             -- the window has been closed or the buffer is somehow no longer valid
             unsubscribe(true)
-            if win_valid then
-                vim.api.nvim_win_close(win_id, true)
-            end
-            if buf_valid then
-                vim.api.nvim_buf_delete(bufnr, { force = true })
-            end
-            redraw_by_winnr[win_id] = nil
+            M.delete_win_buf(win_id, bufnr)
```
kylo252, line 195: I'm not sure why this call is here. Can we even delete a window with an invalid id?
However, I think there's more to it than disabling this function alone.

williamboman: I'd say there's no specific reason other than making sure to do an exhaustive clean up should it ever try to draw/render when the buffer-window pair is invalid to avoid potential lingering windows/buffers.

kylo252: I think I found the bug, please try the latest commit and let me know if it works for you.

williamboman: Hmm it kinda defeats the purpose I think. We need to make sure to delete the buffer (if it's still valid), otherwise you can split it into a new window. The window management is pretty solid I think, thanks to the autocmds and whatnot. This is mostly about ensuring that the buffer follows suite (deletes itself when the popup window closes, or a different way of expressing it - doesn't allow itself to be loaded into another window than the floating window).

What about removing this delete call from the render function (seeing as there is no strong reason to have it in here in the first place), and keeping it as it was (delete both win and buf)? Would that work?

kylo252: Okay, at least I feel like there's more info to go about debugging this now. Here's what happens when you try to list the buffers (notice how a new window **should** wipeout everything). You can also add `previewer=false` to remove the preview warning.

TvazZ90yRF

williamboman: Hmm that's probably just telescope gathering all buffers before opening a new window. When the telescope window is later opened our buffer will have been wiped. If that's the case I feel like it's not related to this plugin at all?

Try doing a vertical split `:vsp` when the floating window is open. Since the buffer will be loaded into a new window before the floating window is closed, it won't delete itself (`bufhidden` behavior is never triggered). The next time it tries to render it will suspend because the original floating window handle is no longer valid. This is why I want the `delete_win_buf` function to delete both the provided window and buffer, because it will be called on `WinLeave`, which will happen in due time. Due to this call to `delete_win_buf` the buffer will actually be deleted, and the newly opened window will be closed.

kylo252: I got stuck trying my hand at this (https://github.com/kylo252/nvim-lsp-installer/tree/display-refactor), but it's way too complicated. I should've started trying to understand how it's used before diving head first into such adventures 😅 

I can't tell you if the current architecture is good or not, mostly because I don't fully get it. Maybe my problem is that I'm treating it as a singleton or that it should be a singleton. I can't help but feel that you have some redundant state management, when you can just instantiate a new object. 

Instead there's a display object that needs to go through this confusing lifecycle: 

🟢 `new()` 
➡️ `view()`
➡️ `init()` "wait, what does `new()` do then? "
➡️ `open()` "open what? what did `view()` do before? how come you can call `view()` before `init()`?"

Some of this might be easily remedied with some renaming, well, I hope so!

ref: https://github.com/williamboman/nvim-lsp-installer/blob/89b5a04fb5bce23cdd77e5b9f2e175f60404e429/lua/nvim-lsp-installer/ui/status-win/init.lua#L236-L268

williamboman: Haha yes it's all a bit thrown together at the moment.

1. Make it work <--- still here :D
1. Make it right
1. Make it fast 

The UI framework I built here basically have 3 main constructs:

1) The stateless "view" function
The view function is responsible for returning an in-memory representation of what should be rendered. This representation is basically just a tree structure of nodes (I believe currently, leaf nodes are either always of the `Ui.Node` or `Ui.HlTextNode` types). It holds no state by itself, but instead receives the state as an argument. This is what's registered when calling `.view(fn)`.

2) The state container
The state container is a simple container for a piece of mutable state. A state container allows for a) mutating the current state, b) retrieving the current state, and c) subscribing to state changes (this is what schedules the next render). The state container is set up when calling `.init(initial_state)`.

3) The "display" object
The display is basically a representation of a buffer-window pair. The window can be any valid neovim window, and I can imagine there being multiple different variants of it (floating, "normal", preview, etc.). What type it is should have no bearing on what can be rendered - but the renderer might do things differently depending on what type it is. The display object is responsible for translating the in-memory representation provided by the view function into `vim.api.nvim_buf` calls (to set lines, highlights, extmarks, etc.). It's basically responsbile for making sure 1) and 2) has a graphical outlet.

williamboman: > I got stuck trying my hand at this (https://github.com/kylo252/nvim-lsp-installer/tree/display-refactor), but it's way too complicated. I should've started trying to understand how it's used before diving head first into such adventures 😅

Haha I think it looks great though :+1:! I'm actually pretty satisfied with how this PR looks like right now - are you still getting the segfaults or should we just go ahead and merge? I've tested on:

- macOS (recent nightly)
- Windows (recent nightly)
- Ubuntu x64 (recent nightly)
- Alpine Linux arm (recent nightly)

And it all works fine!

I'd be more than happy to receive another PR to improve this piece of code. What do you say?

kylo252: > The UI framework I built here basically have 3 main constructs

This will take some time to process, I'm gonna to have to dig into it at some point if you decide to go with the refactor (in a separate PR ofc, it's too much work to jumble it with this one 😄)

> are you still getting the segfaults or should we just go ahead and merge? I've tested on:

Yeah I'm still getting the segfault unfortunately, by basically only pressing `<c-w>` and I've tried it on multiple systems with the same results. I'm confused how you're not encountering this.

williamboman: Do you have any other plugins enabled? Or can you repro on clean install with only this plugin loaded?

kylo252: Damn, you're right! I just tried it with a `minimal_init.lua` and it seems to be okay. 🎉

So something is up with a configuration or a plugin that I have. That should hopefully make debugging much easier.

williamboman: My bet is on https://github.com/folke/which-key.nvim :D

kylo252: It was actually which-key 😢 

I'm trying to replicate this with other plugins. It seems that Telescope knows how to deal with this, whereas Packer sends a wild error message and destroys both windows.

I have a feeling this could be avoided by checking the filetype, but that only works if the info panel was trying to hijack which-key, and not the other way around.

williamboman: Does it still crash with the latest commit ca446fe59b012b8e18e2b6d1aa74a94e37f3585b?

williamboman: 1486b3700c25f771c4bfb01a3400d94cc5a23b3c seems to fix it! Weird one, not gonna bother investigating further haha

kylo252: I was literally going to add a very similar commit, `vim.schedule` is the way!

williamboman (approved):

PR #108, opened 2021-09-29
don't assume bash location

### LINKED ISSUE #107: Failed to spawn process cmd=/bin/bash pid=ENOENT: no such file or directory
OS: NixOS

The bash binary is not present at `/bin/bash` on systems like NixOS and some BSDs.
However, `/bin/sh` is present for all POSIX compliant systems. On standard linux systems, `/bin/sh` is symlinked to either bash or dash or some equivalent. I think, nvim-lsp-installer, should use `/bin/sh` instead of `/bin/bash`.

williamboman:
Resolves #107.

### DIFF
--- a/CUSTOM_SERVERS.md
+++ b/CUSTOM_SERVERS.md
@@ -80,7 +80,7 @@ available installers that are available out of the box.
 
     #### `shell.bash(raw_script: string, opts?: table)`
 
-    Returns an installer that runs the provided `raw_script` as a bash script (`/bin/bash`).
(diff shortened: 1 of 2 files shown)


williamboman:
Hey @phenax - does this work? I just noticed pretty much all shell usage has been replaced via `uv_spawn` calls, so going for `sh` instead of `bash` should be perfectly safe. Will look into that.

phenax:
@williamboman, this worked!

PR #117, opened 2021-10-02
add server filetype

kylo252:
- feat: add support for querying supported filetypes
- feat: display supported filetypes info


kylo252:
The list formatting is independent, so it's fully customizable. I opted for displaying a max number of 3 types, because that list can get pretty long with JS types.

williamboman:
Hey! Hmm I really don't think it's the right place (or generally speaking the right plugin) to show this info. I saw that `:LspInfo` already shows this info, any reason you think it should be shown in the installer info too?

lua/nvim-lsp-installer/server.lua
```diff
@@ -66,6 +67,28 @@ function M.Server:get_default_options()
     return vim.deepcopy(self._default_options)
 end
 
+function M.Server:get_supported_filetypes()
```
williamboman, line 70: At first glance it looks like this would not capture any filetypes that is overridden by users when calling `.setup {}`.

lua/nvim-lsp-installer/server.lua
```diff
@@ -66,6 +67,28 @@ function M.Server:get_default_options()
...
+    log.debug("got filetypes query request for: " .. self.name)
+
+    if self._default_options.filetypes then
+        return self._default_options.filetypes
+    end
+
+    local configs = require "lspconfig/configs"
+    local default_server_config = "lspconfig/" .. self.name
+    pcall(require, default_server_config)
```
williamboman, line 79: I believe by interacting with lspconfig like this, it will cause the server to show up in `:LspInfo` (which has caused some confusion before, especially for servers that aren't installed). Since this would be called for every server in the UI, it'd cause all LSP servers to show up in `:LspInfo`

kylo252:
> Hey! Hmm I really don't think it's the right place (or generally speaking the right plugin) to show this info. I saw that `:LspInfo` already shows this info, any reason you think it should be shown in the installer info too?

This is related to the metadata discussion. I'm not sure about how, or if, you want to display it in the info panel, but I figured it's useful to show you a basic implementation for it. Not all servers have obvious names, e.g. "jedi", "solargraph", etc. It's probably best suited for when you "expand" the information section of a server. 

The other useful part of this PR is the API for figuring out the filetype. If you do a reverse lookup and cache the results, then you can allow for a cheap way to install by filetype: `LspInstall lua`.

lua/nvim-lsp-installer/server.lua
kylo252: I'm not sure where you store that info. My idea was to just avoid doing the lookup more than once.

lua/nvim-lsp-installer/server.lua
kylo252: I'm suspecting this might be the case, which is why I reset `lspconfig/configs` and why I added the note about the possibility of doing this lookup once and caching the values in a file you that renew every once in a while. That might also allow you to remove this list: https://github.com/williamboman/nvim-lsp-installer/blob/ada75fbbad96f63fe5cd7ba3e44d5235ddf33ca6/lua/nvim-lsp-installer/servers/init.lua#L41

Having said that, I don't have any problems with triggering `:LspInfo` after `:LspInstallInfo`, but I've noticed some weird error, about an invalid name, that shows up once or twice, so your YMMV 😄

lua/nvim-lsp-installer/server.lua
williamboman: This plugin doesn't store any information regarding how the servers are set up. The `Server:setup(opts)` function is just a proxy for lspconfig's `setup` functions. Not sure how they store it internally, but I'd imagine there's no public APIs for it

williamboman:
> This is related to the metadata discussion. I'm not sure about how, or if, you want to display it in the info panel, but I figured it's useful to show you a basic implementation for it. Not all servers have obvious names, e.g. "jedi", "solargraph", etc. It's probably best suited for when you "expand" the information section of a server.

Ah yeah that makes sense. I'm slowly making progress on the metadata stuff too. The biggest problem I want to tackle first is how to best orchestrate the collection of metadata seeing as there are quite a few I/O bound operations attached to each server when doing so. At the same time I want to tackle the problem of collecting all outdated servers and providing nice APIs to upgrade these. Once that's done the UI components should be fairly straight forward to implement.

Screenshot 2021-10-02 at 15 57 53

kylo252:
Once that's done the UI components should be fairly straight forward to implement.
> 
> Screenshot 2021-10-02 at 15 57 53

Oh man that looks super nice! 

Not sure if you're using this new function to get the filetypes, but suggested is not the same as supported. The way `lspconfig` works is that it will try to attach the server for **every entry** in the `filetypes`, unless you specifically set it yourself. 

There have been some issues here and there because of some over-/under-reaching server. It's usually best handled upstream. One example would be how you're setting additional filetypes for `graphql` because it wasn't attaching otherwise.

In summary, my idea is to basically show the filetypes primarily for non-installed servers, because that's where most of the confusion lies, as opposed to a server you already have running/installed.

kylo252:
@williamboman, here's the new demo

[asciicast](https://asciinema.org/a/nkziguKTMdKjo0ZD8AsbZVQwJ)

williamboman:
Superseded by #159

PR #118, opened 2021-10-02
fix: Add `--stdio` to the argument of vscode-langservers-extracted

tamago324:
In my environment, it did not start properly without --stdio, so I added the argument.


williamboman (approved):
Ah thanks, there's been some sweeping fundamental changes recently and seems like the arg got lost on the way

tamago324:
Thank you for merging!

PR #138, opened 2021-10-06
Update the go.lua script for the new golang install deprecation

TeoDev1611:
Hi 😃 

I found this install with the old `go get` command binaries with the new version of go this is deprecated

For more information check [here](https://golang.org/doc/go-get-install-deprecation)

### DIFF
--- a/lua/nvim-lsp-installer/installers/go.lua
+++ b/lua/nvim-lsp-installer/installers/go.lua
@@ -26,7 +26,7 @@ function M.packages(packages)
                 pkgs[1] = ("%s@%s"):format(pkgs[1], context.requested_server_version)
             end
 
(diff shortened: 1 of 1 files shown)


nzlov:
Maybe need two different installation scripts.
Or check go version 1.17.

williamboman:
Here's an implementation for fetching textual data from an HTTP endpoint for reference: https://github.com/williamboman/nvim-lsp-installer/blob/4346b3b06103e89ec91e450b29f9a7314dd61495/lua/nvim-lsp-installer/installers/context.lua#L23-L57

williamboman:
I'll close this as the discussion died off, but opened an issue #229 to track this! Happy to receive another PR or reopen this if there are any updates :)

PR #140, opened 2021-10-08
add keybindings to UI window

williamboman:
- Allows for expanding servers to view more information about it.
- Allows for installing/reinstalling/uninstalling servers.

The default keybindings is an attempt to mimic vim-fugitive's :Git
maps, and these can be overriden.

The keybinding implementation in display.lua is a bit hacky, but it
works and the "public" API is at least manageable. This will also
open up for adding more metadata in the future, such as filetype
information, currently installed version, latest available version,
etc.

Also there's Cowth Vader.


williamboman:
https://user-images.githubusercontent.com/6705160/136621980-3765a6a9-9a4b-4753-8e8d-df68a5fc67b4.mov

PR #149, opened 2021-10-13
add eslint, deprecate eslintls

williamboman:


williamboman:
I made the deprecation "flag" a bit richer, to help ease potential future deprecations:

Screenshot 2021-10-14 at 22 53 24

nyngwang:
@williamboman Might I ask you some questions about eslint setup? If you're also using reddit, here is my post: https://www.reddit.com/r/neovim/comments/sd248x/how_to_setup_eslint_and_prettier_for_ts_project/. (The point it is related is that I want to setup this up using your plugin. I won't change)

williamboman:
> When working on a file loaded into a buffer, I can have only that buffer eslint-ed & prettier-ed by a keymapping.

See the following functions:
- `:h vim.lsp.buf.formatting()`
- `:h vim.lsp.buf.formatting_sync()`
- `:h vim.lsp.buf.formatting_seq_sync()`

You can then do a keymap to call whichever formatting function you prefer (`vim.api.nvim_set_keymap("n", "<leader>f", "<cmd>lua vim.lsp.buf.formatting()<CR>")`

> When working on a file loaded into a buffer, I can have only that buffer pre-run/preview the result eslint-ed & prettier-ed by a keymapping.

I don't believe there's a builtin way to do this, shouldn't be too difficult to build from scratch though. Depends on what you mean by pre-run/preview the results.

> When I move my cursor from one buffer to another(including the possibility that it jumps between tabs), a different set of config files are adopted automatically.

Are you talking about buffer-specific configs? Not sure why you'd want to do that? Usually settings are applied on a project-level, and these then get picked up by each LSP server (maybe `:g lspconfig-root-dir` is helpful to read)

nyngwang:
> > When working on a file loaded into a buffer, I can have only that buffer eslint-ed & prettier-ed by a keymapping.
> 
> See the following functions:
> 
> * `:h vim.lsp.buf.formatting()`
> * `:h vim.lsp.buf.formatting_sync()`
> * `:h vim.lsp.buf.formatting_seq_sync()`
> 
> You can then do a keymap to call whichever formatting function you prefer (`vim.api.nvim_set_keymap("n", "<leader>f", "<cmd>lua vim.lsp.buf.formatting()<CR>")`
> 

Would these functions respect the config files `.eslintrc` `.prettierrc`?

kylo252:
> When working on a file loaded into a buffer, I can have only that buffer pre-run/preview the result eslint-ed & prettier-ed by a keymapping.

Assuming you mean seeing the result of the fix, and assuming it's offered by the server, then this might be what you're looking for: https://github.com/weilbith/nvim-code-action-menu

> Would these functions respect the config files `.eslintrc` `.prettierrc` by default

These functions are just telling the server to do the formatting, they should not override the settings. So then it's up to the server to respect those files or not, assuming it can find them in the first-place. The easiest way to know the answer is to just test it :)

nyngwang:
@kylo252 Great thanks for your explanation!

PR #159, opened 2021-10-16
feat: autogen metadata table(s)

kylo252:
Here's the other method that I thought could be useful. Feel free to use any of these however you see fit. I included the script that generated this table as a PoC, but you can completely ignore it.

Some possible workflows:
- Have one global metadata table file (either json or lua): 
already available in the PR
- Have a metadata table file per server (either json or lua):
just make it write metadata to a file in each server's directory
- Embed the filetypes metadata directory based on this table:
a lot of manual work probably, unless we write some sed/awk script to deal with that.

### DIFF
--- a/lua/nvim-lsp-installer.lua
+++ b/lua/nvim-lsp-installer.lua
@@ -1,9 +1,11 @@
+local fs = require "nvim-lsp-installer.fs"
 local notify = require "nvim-lsp-installer.notify"
 local dispatcher = require "nvim-lsp-installer.dispatcher"
 local process = require "nvim-lsp-installer.process"
 local status_win = require "nvim-lsp-installer.ui.status-win"
 local servers = require "nvim-lsp-installer.servers"
 local settings = require "nvim-lsp-installer.settings"
+local log = require "nvim-lsp-installer.log"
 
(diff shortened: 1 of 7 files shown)


kylo252:
> Some reflections on how this should be used initially: I know there are more possibilities with this generated approach (which you have given examples of previously), but in the interest of taking one step at a time maybe we should only consider it as a means to access data that is not feasible to do at runtime? Currently, this would be each server's associated filetype.

This is more or less just a template at the moment that you can shape the way you see fit for your workflow, you can just use the generated data and dump each entry in the respective file and call it a day if you want. But I wanted to show how I retrieved that data in case you wanted to have the script in the CI or even offline to validate the data from time to time.

> Also - I don't see any reason to include the homepage in the generated output. This is already readily available in each server instance, no?

I may have missed to explain that part. So I was thinking it would make sense to collect all the "static" metadata in a single table, so "homepage" would be one and so would version number for example. I got that idea when I was looking into how you did the extraction for VSIX and noticed that *.vsixmanifest file. 

I figured that something like this could be useful to streamline the process, or it could be overkill, I'm not really sure 😄

PR #162, opened 2021-10-16
fix(phpactor): install via git clone instead of composer require

Lctrs:
As stated [here](https://github.com/phpactor/phpactor/issues/1322), `composer require` is not a supported way of installing phpactor.
According to the [doc](https://phpactor.readthedocs.io/en/master/usage/standalone.html), we should rely on `git clone` instead.


Lctrs:
`phpactor` has some plugins to enhance its features. They're not installable through composer but rather via a custom phpactor command. Having phpactor installed through `composer require` break this command.

Lctrs:
I wanted to have this PR merged before opening an issue about how we can integrate these plugins installation in the installer.

More info about installing a phpactor extension can be found here : https://phpactor.readthedocs.io/en/master/extensions.html

PR #163, opened 2021-10-17
Eslint doc for how to resolve eslint in yarn2 repo's

NICHTJ3:
This solves the problem resolving eslint in yarn 2 repositories for use with the eslint lsp.

__NOTE__: I feel there is probably a nicer way to do this but this atleast works


NICHTJ3:
This is the issue upstream that this resolves: https://github.com/microsoft/vscode-eslint/issues/601

NICHTJ3:
I've updated this PR to just add the doc. It's worded pretty poorly so I'm open to criticism 😂

PR #166, opened 2021-10-17
add setting for configuring server install dir

### LINKED ISSUE #164: [Feature] add config to modify SERVERS_ROOT_DIR
Sometimes, I run portable nvim  on the USB disk.
default SERVERS_ROOT_DIR in stdpath("data") is not inconvenient.
Could you add a option to allow modify  this value?

williamboman:
Resolves #164.


williamboman:
@kylo252 👋  For people using LunarVim and who want to change this particular setting - will they be able to provide these settings before any other interactions with this plugin takes place? The reason I ask is that the `install_root_dir` pretty much gets cached all over the place the moment you start interacting with servers (like for example `.get_server(server_name)` or `.on_server_ready()`) - so it's important that settings are provided as early as possible (`install_root_dir` would currently be the only setting where this is important however).

Issue #205, opened 2021-10-28
jdtls lombok


Mupu:
How do you configure this plugin to start it with lombok? I saw it even downloads it. From what I've read you have to pass the javaagent argument but I haven't found a way to do that here. Any ideas?

williamboman:
This should be default behavior now!

PR #212, opened 2021-10-30
Add lombok support #205

s1n7ax:
At the moment lombok is downloaded to the LS directory but never used. This loads the plugin when loading the `jdtls`.

### DIFF
--- a/lua/nvim-lsp-installer/servers/jdtls/init.lua
+++ b/lua/nvim-lsp-installer/servers/jdtls/init.lua
@@ -20,10 +20,9 @@ return function(name, root_dir)
             "-Dlog.level=ALL",
             "-Xms1g",
             "-Xmx2G",
+            "-javaagent:" .. lombok,
             "-jar",
             jar,
-            "-javaagent:" .. lombok,
-            "-Xbootclasspath/a:" .. lombok,
             "-configuration",
             path.concat {
                 root_dir,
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Is the `-Xbootclasspath/a` argument also not needed, for older Java versions?

The reason the jdtls cmd doesn't include Lombok is because it follows [lspconfig's defintion](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/jdtls.lua). We need to override the cmd here because lspconfig hardcodes installation paths.

The idea is that people can include lombok manually should they want it (like [so](https://github.com/williamboman/nvim-lsp-installer/issues/205#issuecomment-954267061)), but at the same time I think it makes sense to use it by default. Do you know whether the lombok agent comes with much overhead?

s1n7ax:
> Is the -Xbootclasspath/a argument also not needed, for older Java versions?

@williamboman Java 9 removed the support for the flag I think. Not quite sure but releases from last 2 year doesn't seem to require the Xbootclasspath flag.

> The reason the jdtls cmd doesn't include

Eclipse and Idea does not include Lombok too. But they are easy to enable in IDEs.

> The idea is that people can include lombok manually should they want it (like so)

It doesn't work for me. Note that the where you put the flag is important here. If you append it after -jar it won't work

> Do you know whether the lombok agent comes with much overhead?

Lombok is only being used at compile time. I don't think it is heavy at all.

williamboman (approved):
Sounds reasonable! I was considering adding a setting for this, but let's run with this

PR #217, opened 2021-10-30
colocate tmpdir with the server's install dir to avoid cross-fs issues

### LINKED ISSUE #215: Every language server fails to install
For example - I run :LspInstall pyright and get this in the log.
Every single language I've tried nets the same result.
I'm using LunarVim
```
[DEBUG Sat 30 Oct 2021 06:21:48 AM EDT] ...-installer/lua/nvim-lsp-installer/ui/status-win/init.lua:612: Installing server {
  _default_options = {
    cmd = { "/home/cs/.local/share/nvim/lsp_servers/python/node_modules/.bin/pyright-langserver", "--stdio" }
  },
  _installer = <function 1>,
  homepage = "https://github.com/microsoft/pyright",
  name = "pyright",
  root_dir = "/home/cs/.local/share/nvim/lsp_servers/python",
  <metatable> = <1>{
    __index = <table 1>,
    get_default_options = <function 2>,
    get_supported_filetypes = <function 3>,
    install = <function 4>,
    install_attached = <function 5>,
    is_installed = <function 6>,
    new = <function 7>,
    setup = <function 8>,
    uninstall = <function 9>
  }
} nil
[INFO  Sat 30 Oct 2021 06:21:48 AM EDT] ...-installer/lua/nvim-lsp-installer/ui/status-win/init.lua:547: Starting install server_name="pyright", requested_version="N/A"
[DEBUG Sat 30 Oct 2021 06:21:48 AM EDT] ...r/start/nvim-lsp-installer/lua/nvim-lsp-installer/fs.lua:56: fs: mkdir /tmp/nvimgOwC9y/8
[DEBUG Sat 30 Oct 2021 06:21:48 AM EDT] ...nvim-lsp-installer/lua/nvim-lsp-installer/ui/display.lua:393: Opening window
[DEBUG Sat 30 Oct 2021 06:21:48 AM EDT] ...rt/nvim-lsp-installer/lua/nvim-lsp-installer/process.lua:116: Spawning cmd="npm", spawn_opts={
  args = { "init", "--yes", "--scope=lsp-installer" },
  cwd = "/tmp/nvimgOwC9y/8"
}
[DEBUG Sat 30 Oct 2021 06:21:48 AM EDT] ...rt/nvim-lsp-installer/lua/nvim-lsp-installer/process.lua:162: Spawned with pid 355124
[DEBUG Sat 30 Oct 2021 06:21:49 AM EDT] ...rt/nvim-lsp-installer/lua/nvim-lsp-installer/process.lua:148: Job pid=355124 exited with exit_code=0, signal=0
[DEBUG Sat 30 Oct 2021 06:21:49 AM EDT] ...rt/nvim-lsp-installer/lua/nvim-lsp-installer/process.lua:116: Spawning cmd="npm", spawn_opts={
  args = { "install", "pyright" },
  cwd = "/tmp/nvimgOwC9y/8"
}
[DEBUG Sat 30 Oct 2021 06:21:49 AM EDT] ...rt/nvim-lsp-installer/lua/nvim-lsp-installer/process.lua:162: Spawned with pid 355135
[DEBUG Sat 30 Oct 2021 06:21:49 AM EDT] ...rt/nvim-lsp-installer/lua/nvim-lsp-installer/process.lua:148: Job pid=355135 exited with exit_code=0, signal=0
[DEBUG Sat 30 Oct 2021 06:21:49 AM EDT] ...r/start/nvim-lsp-installer/lua/nvim-lsp-installer/fs.lua:38: fs: rename /tmp/nvimgOwC9y/8 /home/cs/.local/share/nvim/lsp_servers/python
[INFO  Sat 30 Oct 2021 06:21:49 AM EDT] ...-installer/lua/nvim-lsp-installer/ui/status-win/init.lua:566: Installation completed server_name="pyright", success=false
```

williamboman:
Fixes #215.

### DIFF
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -99,7 +99,7 @@ function M.Server:_setup_install_context(context)
     fs.rm_mkdirp(context.install_dir)
 
     if not fs.dir_exists(settings.current.install_root_dir) then
-        fs.rm_mkdirp(settings.current.install_root_dir)
+        fs.mkdirp(settings.current.install_root_dir)
     end
 end
 
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -126,6 +126,7 @@ function M.Server:install_attached(context, callback)
                         return
                     end
                 end
+                fs.mkdir(self.root_dir)
                 local rename_ok, rename_err = pcall(fs.rename, context.install_dir, self.root_dir)
                 if rename_ok then
                     vim.schedule(function()
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -5,6 +5,7 @@ local settings = require "nvim-lsp-installer.settings"
 local installers = require "nvim-lsp-installer.installers"
 local servers = require "nvim-lsp-installer.servers"
 local status_win = require "nvim-lsp-installer.ui.status-win"
+local path = require "nvim-lsp-installer.path"
 
 local M = {}
 
@@ -93,26 +94,21 @@ function M.Server:install()
 end
 
 ---@param context ServerInstallContext
+function M.Server:_setup_install_context(context)
+    context.install_dir = path.concat { settings.current.install_root_dir, ("%s.tmp"):format(self.name) }
+    fs.rm_mkdirp(context.install_dir)
+
+    if not fs.dir_exists(settings.current.install_root_dir) then
+        fs.rm_mkdirp(settings.current.install_root_dir)
+    end
+end
+
+---@param context ServerInstallContext
 ---@param callback ServerInstallCallback
 function M.Server:install_attached(context, callback)
-    ---@param path string
-    local function mkdir(path)
-        local mkdir_ok, mkdir_err = pcall(fs.mkdir, path)
-        if not mkdir_ok then
-            log.fmt_error("Failed to mkdir. path=%s error=%s", path, mkdir_err)
-            context.stdio_sink.stderr(("Failed to create directory %q.\n"):format(path))
-            context.stdio_sink.stderr(tostring(mkdir_err) .. "\n")
-            return false
-        end
-        return true
-    end
-
-    context.install_dir = vim.fn.tempname()
-    if not mkdir(context.install_dir) then
-        callback(false)
-        return
-    end
-    if not fs.dir_exists(settings.current.install_root_dir) and not mkdir(settings.current.install_root_dir) then
+    local context_ok, context_err = pcall(self._setup_install_context, self, context)
+    if not context_ok then
+        log.error("Failed to setup installation context.", context_err)
         callback(false)
         return
     end
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -117,22 +117,25 @@ function M.Server:install_attached(context, callback)
         self,
         vim.schedule_wrap(function(success)
             if success then
-                if fs.dir_exists(self.root_dir) then
-                    local rmrf_ok, rmrf_err = pcall(fs.rmrf, self.root_dir)
-                    if not rmrf_ok then
-                        log.fmt_error("Failed to rmrf. path=%s error=%s", self.root_dir, rmrf_err)
-                        context.stdio_sink.stderr "Failed to delete existing installation.\n"
-                        context.stdio_sink.stderr(tostring(rmrf_err) .. "\n")
-                        return
-                    end
+                -- 1. Remove and recreate final installation directory
+                local rmrf_ok, rmrf_err = pcall(fs.rm_mkdirp, self.root_dir)
+                if not rmrf_ok then
+                    log.fmt_error("Failed to rm_mkdirp. path=%s error=%s", self.root_dir, rmrf_err)
+                    context.stdio_sink.stderr "Failed to remove and recreate final installation directory.\n"
+                    context.stdio_sink.stderr(tostring(rmrf_err) .. "\n")
+                    callback(false)
+                    return
                 end
-                fs.mkdir(self.root_dir)
+
+                -- 2. Move the temporary install dir to the final installation directory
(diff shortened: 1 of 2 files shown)


kylo252 (approved):

kylo252:
Tested and works for LunarVim 🥳

lua/nvim-lsp-installer/server.lua
```diff
@@ -93,26 +94,21 @@ function M.Server:install()
...
+function M.Server:_setup_install_context(context)
+    context.install_dir = path.concat { settings.current.install_root_dir, ("%s.tmp"):format(self.name) }
+    fs.rm_mkdirp(context.install_dir)
 
-    context.install_dir = vim.fn.tempname()
-    if not mkdir(context.install_dir) then
-        callback(false)
-        return
+    if not fs.dir_exists(settings.current.install_root_dir) then
+        fs.rm_mkdirp(settings.current.install_root_dir)
```
kylo252, line 102: I know that this general bait-and-switch works for this, but I would really suggest to reconsider if it's causing problems. 

Is this option relevant for installlations going through managers? As I understand it, it only matters for curl and uncompress downloads. In which case there are easier ways to handles this natively, even if it meant creating a different handler for Windows.

williamboman: This will make sure that the root install directory (`~/.local/share/nvim/lsp_servers`) exists. This is needed for all server installers as they all write to the local directory, no matter if it's calling another package manager or just curl et al. The only files that should be expected to be written outside of `~/.local/share/nvim/lsp_servers` are caches, if any utility caches things

williamboman: Generally this code branch will only execute the first time you install a server, or after you've `:LspUninstallAll`

kylo252: I'll try to put this into two contexts:
1- managers-supported servers: you don't need a tempdir if the original name is not valid, whatever the reason for that may be, because you can either aquire the name from the beginning or not. You can always ask the manager afterwards to confirm the installtion. 
```bash
npm list # or pip list, or something similar
```
2- "manual" servers: you use the tempdir in case the downloader gets interrupted, and because that's how you know that the installation completed.

- unix-ish

```shell
# you can also use a tempfile instead
DOWNLOAD_DIR="$(mktemp -d)"
curl -L "$RELEASE_URL" --remote-name --output-dir "$DOWNLOAD_DIR"
tar -xzf "$DOWNLOAD_DIR/$ARCHIVE_NAME.tar.gz" --directory "$TARGET_DIR"
```

- windows

```powershell
$tmp = New-TemporaryFile
iwr -UseBasicParsing -Uri $ReleaseUrl -OutFile $tmp
Expand-Archive -Path $tmp -DestinationPath $TargetDir
```

williamboman: > 1- managers-supported servers: you don't need a tempdir if the original name is not valid, whatever the reason for that may be, because you can either aquire the name from the beginning or not. You can always ask the manager afterwards to confirm the installtion.

I'm not following you here at all 🤔 . Which name are you referring to? Do you mean hooking up `:is_installed()` to the underlying package manager? I'd really like to minimize dependency on these package managers as much as possible because 1) all interactions are I/O bound, 2) would have to support all the different versions of said package manager, and be able to parse all different versions of its outputs.

Using a tmpdir as a proxy for determining whether a server was successfully installed, no matter if it's installed using a pkg manager or not, I think is a very effective but more importantly super simple method.

> 2- "manual" servers: you use the tempdir in case the downloader gets interrupted, and because that's how you know that the installation completed.

This is sort of how the `main` branch works. It uses `tempname()` to create a temporary directory, but this is causing issues when /tmp is mounted as a separate file system. I think your examples are pretty contrived, installers are often not that simple.


An "installer" is really just a function with the signature `function (server, callback, context)`. These functions can be composed together, and you can write higher-order installers that allow for control flow based on some logic and whatnot. There's one important property with an installer function, and that is that whenever they interact with the filesystem, they must do so using relative paths. This allows for easier governance in terms of where files and directories actually end up (we use this to run everything in the tmpdir, and when all installers succeed we move it to the final dir). I don't think changing to a system where we run command 1, 2, 3, 4 in a tmpdir, and on the 5th command we need to run a command that compiles the previous steps and writes to the final install dir. That would make composing installers nearly impossible and every server installer would pretty much end up being siloed

williamboman: I'll go ahead and merge as this seems to fix the issue, so it can't be worse than in `main` at least :D

kylo252: I wasn't really talking about the `is_installed()`, rather the confirmation of "install complete".

> I think is a very effective but more importantly super simple method.

I'm 100% with you on that, which is why I prefaced this entire discussion with "but I would really suggest to reconsider _if it's causing problems_."

>  I think your examples are pretty contrived, installers are often not that simple.

I'm specifically targeting the installers relying on: https://github.com/williamboman/nvim-lsp-installer/blob/868ff4237e2a63ebfd43ef34bab6ea9e2421031b/lua/nvim-lsp-installer/installers/std.lua#L9-L12

It is not really doing anything that isn't covered by my example. You can either tell the downloader to return the path of the tempfile, or make the uncompress a callback negating the need for the tempfile altoghether. I feel like this is somehow a standard way of doing things.

williamboman: `download_file` is being composed by a bunch of other installers though, for example [`unzip_remote`](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/installers/std.lua#L56-L63). Since it just writes the file relative to the `context.install_dir`, it makes composition possible. This also gets a bit more complicated when you're downloading more than one file, for example [in solang](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/solang/init.lua#L12-L45)

kylo252: Let me know if you want to extract those system calls into separate script modules `download_file.(sh|ps1)`. They can just accept a few input paramters, so it's a drop in replacement for those calls. It's probably more relevant for PowerShell than the others.

PR #220, opened 2021-10-31
server: add attach_buffers() method

### LINKED ISSUE #188: conflict with last-position-jump
I add last-position-jump in my config 
```
				*restore-cursor* *last-position-jump*  >
    autocmd BufReadPost *
      \ if line("'\"") >= 1 && line("'\"") <= line("$") && &ft !~# 'commit'
      \ |   exe "normal! g`\""
      \ | endif
```
after I add `vim.cmd [[ do User LspAttachBuffers ]]` for `nvim-lsp-installer`, It can't work.

Is there any way to avoid this problem?

config
```
lsp_installer.on_server_ready(function(server)
    local opts = {}
    server:setup(opts)
    vim.cmd [[ do User LspAttachBuffers ]]
end)
```

Reproduce Step:
```
comment vim.cmd [[ do User LspAttachBuffers ]]
nvim test.txt +6  --jump file line 6

uncomment vim.cmd [[ do User LspAttachBuffers ]]
nvim test.txt +6  --can't jump file line 6
```

williamboman:
Resolves #188.

### DIFF
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -46,7 +46,7 @@ function M.Server:new(opts)
 end
 
 ---@param opts table @User-defined options. This is directly passed to the lspconfig's setup() method.
-function M.Server:setup(opts)
+function M.Server:lspconfig_setup(opts)
     if self._pre_setup then
         log.fmt_debug("Calling pre_setup for server=%s", self.name)
         self._pre_setup()
@@ -68,6 +68,12 @@ function M.Server:setup(opts)
             ):format(self.name)
         )
     end
+end
+
+---@param opts table @User-defined options. This is directly passed to the lspconfig's setup() method.
+function M.Server:setup(opts)
+    self:lspconfig_setup(opts)
+    self:attach_buffers()
 end
 
 ---Attaches this server to all current open buffers with a 'filetype' that matches the server's configured filetypes.
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -84,7 +84,7 @@ function M.Server:attach_buffers()
         log.fmt_trace("Attaching server=%s to bufnr=%s", self.name, bufnr)
         lspserver.manager.try_add_wrapper(bufnr)
     end
-    log.debug("lspconfig.manager.try_add_wrapper() successfully called", self.name)
+    log.debug("Successfully attached server to buffers", self.name)
 end
 
 ---Registers a handler (callback) to be executed when the server is ready to be setup.
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -45,8 +45,9 @@ function M.Server:new(opts)
     }, M.Server)
 end
 
----@param opts table @User-defined options. This is directly passed to the lspconfig's setup() method.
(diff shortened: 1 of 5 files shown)


lua/nvim-lsp-installer/server.lua
```diff
@@ -70,6 +70,17 @@ function M.Server:setup(opts)
...
+function M.Server:attach_buffers()
+    log.debug("Attaching server to buffers", self.name)
+    local lspserver = require("lspconfig")[self.name]
+    for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
+        log.fmt_trace("Attaching server=%s to bufnr=%s", self.name, bufnr)
+        lspserver.manager.try_add_wrapper(bufnr)
+    end
+    log.debug("lspconfig.manager.try_add_wrapper() successfully called", self.name)
+end
+
```
williamboman, line 83: @mjlbach :wave: would you say this is correct usage of the server manager? The reason this is needed is because a server may become available after neovim initialization (for example installing a new server), and so we want to attach any open buffers to it

mjlbach: I think its fine

lua/nvim-lsp-installer/server.lua
```diff
@@ -70,6 +70,23 @@ function M.Server:setup(opts)
     end
 end
 
+---@param opts table @User-defined options. This is directly passed to the lspconfig's setup() method.
+function M.Server:setup(opts)
+    self:lspconfig_setup(opts)
+    self:attach_buffers()
```
williamboman, line 76: @kylo252 Will this be problematic for your integration? The `:setup()` method will attach buffers by default now, gives a nicer public API and allows to easily deprecate the `User LspAttachBuffers` autocmd

kylo252: I was actually considering pushing this upstream since I made the changes in [`546ece0` (#1863)](https://github.com/LunarVim/LunarVim/pull/1863/commits/546ece0b58c6ab0bad24d8092f236fe7ca7b2b9b), but I made a mistake there and didn't check for for the filetype (`try_add` vs `try_add_wrapper`), which you already fixed here 😄

lua/nvim-lsp-installer/server.lua
```diff
@@ -70,6 +71,24 @@ function M.Server:setup(opts)
...
+    self:setup_lsp(opts)
+    self:attach_buffers()
+end
+
+---Attaches this server to all current open buffers with a 'filetype' that matches the server's configured filetypes.
+function M.Server:attach_buffers()
+    log.debug("Attaching server to buffers", self.name)
+    local lsp_server = require("lspconfig")[self.name]
+    for _, bufnr in ipairs(vim.api.nvim_list_bufs()) do
+        log.fmt_trace("Attaching server=%s to bufnr=%s", self.name, bufnr)
```
kylo252, line 86:
```suggestion
        log.fmt_debug("Attaching server=%s to bufnr=%s", self.name, bufnr)
```

kylo252 (approved):

lua/nvim-lsp-installer/server.lua
williamboman: I think that might add too much noise. I think it's fine to assume vim.api.nvim_list_bufs() works as it should

PR #226, opened 2021-10-31
Attempt to fix solargraph install

elken:
Needs more testing from other users, but before this nothing works on the LSP end for me.

### DIFF
--- a/lua/nvim-lsp-installer/servers/solargraph/init.lua
+++ b/lua/nvim-lsp-installer/servers/solargraph/init.lua
@@ -6,7 +6,10 @@ return function(name, root_dir)
         name = name,
         root_dir = root_dir,
         homepage = "https://solargraph.org",
-        installer = gem.packages { "solargraph" },
+        installer = {
+            gem.packages { "irb" },
+            gem.packages { "solargraph" },
+        },
         default_options = {
             cmd = { gem.executable(root_dir, "solargraph"), "stdio" },
             cmd_env = gem.env(root_dir),


lua/nvim-lsp-installer/servers/solargraph/init.lua
```diff
@@ -6,7 +6,10 @@ return function(name, root_dir)
         name = name,
         root_dir = root_dir,
         homepage = "https://solargraph.org",
-        installer = gem.packages { "solargraph" },
+        installer = {
+            gem.packages { "irb" },
+            gem.packages { "solargraph" },
+        },
```
williamboman, line 12:
```suggestion
        installer = gem.packages { "solargraph", "irb" },
```

williamboman (commented):
Thanks for opening a PR! What exactly is this trying to fix? Why does adding irb fix it?

elken:
Not sure why it was complaining in the first place, but on my end it was lacking `irb/notifier`.

RE: your proposed change, I tried that and was getting `gem` complaining about a malformed URI.

williamboman:
> Not sure why it was complaining in the first place, but on my end it was lacking `irb/notifier`.
> 
> RE: your proposed change, I tried that first and was getting `gem` complaining about a malformed URI.

How exactly does the issue manifest itself? I've looked through docs and sources and can't find anything that explicitly states `irb` as a dependency.

elken:
It seems like it was a dependency but was removed as one. I had the sane issue across 2 systems, but now I'm suspicious so I'll do more digging on more projects to get a better sample size.

williamboman:
Hey @elken, are you still struggling with Solargraph? I seem to always have issues with Ruby, one way or another, so I can't really test it atm, but I remember Solargraph was working when it was added.

elken:
Thank you for reminding me :) I don't work with Ruby so I haven't had a chance to test, I've left myself a TODO to try some ruby projects in the next few days.

williamboman:
> Will try in isolation too, because I'm suspicious my environment might interfere (although you saying you're having issues too makes me more confident)

Haha yeah, my Ruby installations always end up so unreliable and keep breaking for all various kinds of reasons. Here's my error, FWIW:

```
        Building native extensions. This could take a while...
        ERROR:  Error installing irb:
        	ERROR: Failed to build gem native extension.
        
            current directory: /Users/williamboman/.local/share/nvim/lsp_servers/solargraph.tmp/gems/io-console-0.5.9/ext/io/console
        /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/bin/ruby -I /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0 -r ./siteconf20211129-89237-dutm3x.rb extconf.rb
        *** extconf.rb failed ***
        Could not create Makefile due to some reason, probably lack of necessary
        libraries and/or headers.  Check the mkmf.log file for more details.  You may
        need configuration options.
        
        Provided configuration options:
        	--with-opt-dir
        	--without-opt-dir
        	--with-opt-include
        	--without-opt-include=${opt-dir}/include
        	--with-opt-lib
        	--without-opt-lib=${opt-dir}/lib
        	--with-make-prog
        	--without-make-prog
        	--srcdir=.
        	--curdir
        	--ruby=/System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/bin/$(RUBY_BASE_NAME)
        /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/mkmf.rb:467:in `try_do': The compiler failed to generate an executable file. (RuntimeError)
        You have to install development tools first.
        	from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/mkmf.rb:585:in `block in try_compile'
        	from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/mkmf.rb:534:in `with_werror'
        	from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/mkmf.rb:585:in `try_compile'
        	from /System/Library/Frameworks/Ruby.framework/Versions/2.6/usr/lib/ruby/2.6.0/mkmf.rb:852:in `macro_defined?'
        	from extconf.rb:7:in `<main>'
        
        To see why this extension failed to compile, please check the mkmf.log which can be found here:
        
          /Users/williamboman/.local/share/nvim/lsp_servers/solargraph.tmp/extensions/universal-darwin-21/2.6.0/io-console-0.5.9/mkmf.log
        
        extconf failed, exit code 1
        
        Gem files will remain installed in /Users/williamboman/.local/share/nvim/lsp_servers/solargraph.tmp/gems/io-console-0.5.9 for inspection.
        Results logged to /Users/williamboman/.local/share/nvim/lsp_servers/solargraph.tmp/extensions/universal-darwin-21/2.6.0/io-console-0.5.9/gem_make.out
```

elken:
That looks like [this](https://stackoverflow.com/questions/46377667/docker-for-mac-mkmf-rb-cant-find-header-files-for-ruby) maybe? I hate how xcode gets in the way...

williamboman:
Yeah I've given up trying to make Ruby work on mac. It's so prone to breakage. I don't use it so it's not worth the effort anymore.

elken:
I don't blame you. I think the distinction for this is we make sure everything is installed and runnable, any _actual_ issues are for the LSP :)

williamboman:
On an Ubuntu 18.04 machine I manage to install and run solargraph just fine via nvim-lsp-installer though

elken:
IIRC I had the same issue on Fedora 35 that I did on macOS, but I will work on creating as isolated of a case as I can.

elken:
Just tested in a fresh toolbox, installed `ruby ruby-devel neovim clang llvm` to get a basic working setup, opened the same  project I've been using and it errors.

It suggests it is indeed isolated to the project, and so I don't contradict what I said; I consider it to be a bug with the server so I'm going to close this as I don't work with ruby so don't have enough time to find a more correct fix.

elken:
Yup that's the same issue I got but I couldn't replicate it at the time. Might be one to reopen @williamboman ?

> This does not

Read the conversation

PR #228, opened 2021-10-31
respect autostart setting

### LINKED ISSUE #227: New attach_buffers command not respecting autostart
With commit 14b64a2, servers that I have configured with `autostart=false` are being attached to buffers without being explicitly started (the `autostart` option seems to be ignored). If I revert to d5f370b and use `vim.cmd('do User LspAttachBuffers')`, behavior is as expected (`autostart=false` servers aren't attached until explicitly started).

To reproduce, configure an lsp-installer managed language server with `autostart=false`, open a matching filetype, and observe that the language server is started.

I also observed with commit 14b64a2 that a FileType autocommand I had setup to manually start one language server no longer appears to be firing. Again, switching back to d5f370b restores the expected behavior (the command fires and manually starts the language server).

williamboman:
Fixes #227.

### DIFF
--- a/lua/nvim-lsp-installer/server.lua
+++ b/lua/nvim-lsp-installer/server.lua
@@ -75,7 +75,9 @@ end
 ---@param opts table @The lspconfig server configuration.
 function M.Server:setup(opts)
     self:setup_lsp(opts)
-    self:attach_buffers()
+    if not (opts.autostart == false) then
+        self:attach_buffers()
+    end
 end
 
 ---Attaches this server to all current open buffers with a 'filetype' that matches the server's configured filetypes.


lua/nvim-lsp-installer/server.lua
```diff
@@ -75,7 +75,9 @@ end
 ---@param opts table @The lspconfig server configuration.
 function M.Server:setup(opts)
     self:setup_lsp(opts)
-    self:attach_buffers()
+    if not (opts.autostart == false) then
+        self:attach_buffers()
+    end
```
williamboman, line 80: @mjlbach :wave: what do you think about doing this in lspconfig by default (attaching all open buffers when calling `.setup()`)? Happy to open PR

PR #233, opened 2021-11-01
feat: server discoverability by filetype

### LINKED ISSUE #177: [RFC] More discoverable identifiers for the :LspInstall, :LspUninstall commands
## Background
The `:LspInstall` and `:LspUninstall` commands currently only allow providing the lspconfig server names. This can be somewhat confusing for people who have not spent any time on LSP development - for example a Python LSP can be installed via `:LspInstall pyright`, `:LspInstall pylsp`, and `:LspInstall jedi_language_server`. Unless you are familiar with these Python LSPs it's 1) difficult to discover these in the first place (you may get lucky by typing `py<Tab>`), and 2) knowing which one to choose.

### 1. Prefix server names with a more discoverable name

Each server may be associated with one or more constructs that are more accessible to the average user. Arguably, the dominant construct in this context would be which language/framework/library the LSP targets. This mapping is already provided in the [README](https://github.com/williamboman/nvim-lsp-installer#available-lsps).

These prefixes will only be available for autocompletion in the command line. Calls to the underlying Lua APIs would not respect the prefix, and will return error should a prefix be included. In order to reduce noise, the server's true names will no longer be available for autocompletion in the command line - but they will still be considered valid inputs.

**Examples**
- `:LspInstall python:pylsp@some-version`
- `:LspInstall python:pylsp`
- `:LspInstall python:pyright`
- `:LspInstall vue:volar`

**Considerations to be made**
- Should there be a limit to how many prefixes a server may have? Are there any cases where it makes sense to have more than one?
- Should autocompletion in the command line still include the server's true names?

-----

### 2. Introduce a new :LspInstallLanguage command

A new `:LspInstallLanguage` command could allow us to have two neatly separated commands for installing a server by using a different identifier. We could also use this command to impose a recommended, default, server per language. In the case where there are more than one LSP per language, this recommended server could either be automatically chosen, or we could prompt the user which one to install (and have the prompt default to the recommended one).

**Examples**
- `:LspInstallLanguage python`
- `:LspInstallLanguage graphql`
- `:LspInstallLanguage vue`

With prompt:
```
:LspInstallLanguage python
Please choose which server you want to install.
(1) pyright (recommended), [2] pylsp, (3) jedi_language_server:
```

We could also use the prompt to signal servers that are unstable, for example if the server has a questionable implementation of the LSP spec. This could potentially help managing expectations and reducing volume of related support issues.

```
:LspInstallLanguage csharp
Please choose which server you want to install.
(1) omnisharp (unstable)
```

## Resolve by
25th October 2021

kylo252:
allow installing per any filetype that does **not** begin with the same 3 letters with the name of its server.

Fixes #177 
Related #232

### DIFF
--- a/lua/nvim-lsp-installer/servers/init.lua
+++ b/lua/nvim-lsp-installer/servers/init.lua
@@ -2,6 +2,7 @@ local Data = require "nvim-lsp-installer.data"
 local path = require "nvim-lsp-installer.path"
 local fs = require "nvim-lsp-installer.fs"
 local settings = require "nvim-lsp-installer.settings"
+local servers_metadata = require "nvim-lsp-installer._generated.metadata"
 
 local M = {}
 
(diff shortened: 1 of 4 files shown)


lua/nvim-lsp-installer/servers/init.lua
```diff
@@ -31,72 +32,7 @@ local INSTALL_DIRS = {
...
-    "tflint",
-    "tsserver",
-    "vala_ls",
-    "vimls",
-    "volar",
-    "vuels",
-    "yamlls",
-    "zls",
-}
+local CORE_SERVERS = Data.set_of(vim.tbl_keys(servers_metadata))
```
williamboman, line 35: This looks interesting, but how will it pick up new servers? Won't it just be permanently based off of what the last hard-coded list of `CORE_SERVERS` was?

kylo252: Yes, but the original list was always hard-coded anyway, but now it lives somewhere else. While with this change, you can get the list of supported servers (somehow) from lspconfig and cross-reference it against what's available. It could also be a scheduled CI job that takes care of that.

This already makes it easier to maintain, and makes sure you only have one source of truth.

williamboman:
Closing in favor for #232, thanks for putting forth this idea though @kylo252!

PR #242, opened 2021-11-04
moved add-opens and add-modules  java options before jar option

ankurs47:
The issue is `--add-opens` and `--add-modules` are arguments to jvm. If we put them after the `-jar` option, they are passed to the java program, not the jvm. Similar fix was done in https://github.com/kabouzeid/nvim-lspinstall/pull/117


williamboman (approved):
Thanks! Let's patch this in lspconfig too, as this cmd is largely a mirror of it (Lombok should be the only diff)

PR #249, opened 2021-11-08
fix: solang needs to specify a target now, can be: ewasm, solana, substrate

ChristianChiarulli:


lua/nvim-lsp-installer/servers/solang/init.lua
```diff
@@ -54,7 +54,7 @@ return function(name, root_dir)
             llvm_installer,
         },
         default_options = {
-            cmd = { path.concat { root_dir, "solang" }, "--language-server" },
+            cmd = { path.concat { root_dir, "solang" }, "--language-server", "--target", "ewasm" },
```
williamboman, line 57: Is `ewasm` some kind universal target? I feel like it'd be best if solang allowed configuring this through project-local config files or similar - gets a bit nasty if someone wants to change this

williamboman (approved):
LGTM! Let's port this patch to lspconfig too?

PR #256, opened 2021-11-11
Get the python executable from settings instead of hard-coding it

tbttfox:
Rather than hard-coding `python` or `python3`, just load the python3 that nvim knows about and let users override this setting. Since users probably already pass the checks in `:checkhealth` (ie. the python3 provider), this shouldn't be a big issue.

I'm stuck with python 2.7 on my machine (Blame Autodesk not updating their embedded python until *AFTER* python 2 was deprecated) so being able to explicitly set the executable path means I won't get stuck running Python 2 with the `python` call

### DIFF
--- a/lua/nvim-lsp-installer/installers/pip3.lua
+++ b/lua/nvim-lsp-installer/installers/pip3.lua
@@ -46,9 +46,13 @@ end
 
 ---@param packages string[] @The pip packages to install. The first item in this list will be the recipient of the server version, should the user request a specific one.
 function M.packages(packages)
-    local py3_host_prog = settings.current.python3_exeutable
+    local py3 = create_installer("python3", packages)
+    local py = create_installer("python", packages)
+    local py3_host_prog = vim.api.nvim_get_var "python3_host_prog"
+
     return installers.pipe {
         context.promote_install_dir(),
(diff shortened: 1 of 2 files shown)


lua/nvim-lsp-installer/installers/pip3.lua
```diff
@@ -46,11 +46,9 @@ end
 
 ---@param packages string[] @The pip packages to install. The first item in this list will be the recipient of the server version, should the user request a specific one.
 function M.packages(packages)
-    local py3 = create_installer("python3", packages)
-    local py = create_installer("python", packages)
     return installers.pipe {
         context.promote_install_dir(),
-        installers.first_successful(platform.is_win and { py, py3 } or { py3, py }), -- see https://github.com/williamboman/nvim-lsp-installer/issues/128
+        create_installer(settings.current.python3_exeutable, packages),
```
williamboman, line 51:
```suggestion
         installers.first_successful(platform.is_win and { py3_host_prog, py, py3 } or { py3_host_prog, py3, py }),
```

williamboman (commented):
Hello! I think it makes sense to source that var to get the executable. I recall considering it in the past but decided not to, can't remember exactly why.

However, far from everyone will have that var defined, so we'll need to keep the current behavior and simply extend it instead

PR #260, opened 2021-11-12
Update lemminx to 0.18.1-451 to resolve crash with maven files

snaggen:
lemminx crashed when I tried to open a pom.xml with version 0.18.0-400, updating to 0.18.1-451 resolved the issue. I had this issue on Fedora 35, just for info.

### DIFF
--- a/lua/nvim-lsp-installer/servers/lemminx/init.lua
+++ b/lua/nvim-lsp-installer/servers/lemminx/init.lua
@@ -31,7 +31,7 @@ return function(name, root_dir)
                 end
             end,
             context.set(function(ctx)
-                ctx.requested_server_version = coalesce(ctx.requested_server_version, "0.18.0-400")
+                ctx.requested_server_version = coalesce(ctx.requested_server_version, "0.18.1-451")
             end),
             context.capture(function(ctx)
                 return std.unzip_remote(
--- a/lua/nvim-lsp-installer/servers/lemminx/init.lua
+++ b/lua/nvim-lsp-installer/servers/lemminx/init.lua
@@ -31,11 +31,11 @@ return function(name, root_dir)
                 end
             end,
             context.set(function(ctx)
-                ctx.requested_server_version = coalesce(ctx.requested_server_version, "0.18.1-451")
+                ctx.requested_server_version = coalesce(ctx.requested_server_version, "LATEST")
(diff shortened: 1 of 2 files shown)


lua/nvim-lsp-installer/servers/lemminx/init.lua
```diff
@@ -31,7 +31,7 @@ return function(name, root_dir)
                 end
             end,
             context.set(function(ctx)
-                ctx.requested_server_version = coalesce(ctx.requested_server_version, "0.18.0-400")
+                ctx.requested_server_version = coalesce(ctx.requested_server_version, "0.18.1-451")
```
williamboman, line 34: I did some investigation and it seems like they also distribute the binary through https://download.jboss.org/jbosstools/vscode/snapshots/lemminx-binary, where there's a convenient "LATEST" entry which we could default to. What about:

```suggestion
                ctx.requested_server_version = coalesce(ctx.requested_server_version, "LATEST")
```

and then further down change the root url to 

```diff
-                    ("https://download.jboss.org/jbosstools/vscode/stable/lemminx-binary/%s/%s.zip"):format(
+                    ("https://download.jboss.org/jbosstools/vscode/snapshots/lemminx-binary/%s/%s.zip"):format(
```

snaggen: I don't have enough experience with lemminx to determine how stable their snapshots tend to be... but if they tend to be stable I assume LATEST would be easier to avoid having to update on every release

williamboman: Yeah pinning versions here is avoided as much as possible (I think it's only done in one other server). It's still 0-version so I guess stability is a subjective matter.

snaggen: I leave it up to you to do as you like, I don't have any strong preferences here.

williamboman: Let's go with using the snapshots distribution channel with LATEST as default

snaggen: I have update the PR to use LATEST

williamboman (approved):

PR #261, opened 2021-11-15
angularls: Add typescript dependency

ilyasakin:
`@angular/language-server` package needs the `typescript` package to be present.

Without the `typescript` package, it outputs following log message:
```
[ERROR][2021-11-15 11:39:05] .../vim/lsp/rpc.lua:462	"rpc"	"/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/.bin/ngserver"	"stderr"	"/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/@angular/language-server/index.js:219\n      throw new Error(`Failed to resolve '${packageName}' with minimum version '${minVersion}' from ` + JSON.stringify(probeLocations, null, 2));\n      ^\n\nError: Failed to resolve 'typescript/lib/tsserverlibrary' with minimum version '4.3' from [\n  \"/home/ashnwor/.local/share/nvim/lsp_servers/angularls\"\n]\n    at resolveWithMinVersion (/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/@angular/language-server/index.js:219:13)\n    at resolveTsServer (/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/@angular/language-server/index.js:228:14)\n    at requireOverride (/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/@angular/language-server/index.js:297:57)\n    at dist/server/logger.js (/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/@angular/language-server/index.js:414:27)\n    at __require (/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/@angular/language-server/index.js:313:44)\n    at Object.<anonymous> (/home/ashnwor/.local/share/nvim/lsp_servers/angularls/node_modules/@angular/language-server/index.js:4557:16)\n    at Module._compile (internal/modules/cjs/loader.js:1085:14)\n    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1114:10)\n    at Module.load (internal/modules/cjs/loader.js:950:32)\n    at Function.Module._load (internal/modules/cjs/loader.js:790:12)\n"
```

then exits with status code 1.

PR #266, opened 2021-11-18
angularls: use server root dir for probe locations

williamboman:


rybertm:
Hi, I'm having a `cmd ["ngserver"] is not executable` error, even though it is installed and I can run it directly from my shell. Maybe an issue with the folder having aliases instead of the actual binary?

Sorry I'm commenting too late, I was trying to debug it and you merged 2 min before I went to report it.

lucasvianav:
> Hi, I'm having a `cmd ["ngserver"] is not executable` error, even though it is installed and I can run it directly from my shell. Maybe an issue with the folder having aliases instead of the actual binary?
> 
> Sorry I'm commenting too late, I was trying to debug it and you merged 2 min before I went to report it, and I skimmed the code, didn't really try it before commenting on the [original PR](https://github.com/williamboman/nvim-lsp-installer/pull/261)

@williamboman I'm having this problem as well. 

Before I had it set up with
```lua
  if server.name == 'angularls' then
    local cmd = {
      server.root_dir .. '/node_modules/.bin/ngserver',
      '--stdio',
      '--tsProbeLocations',
      server.root_dir,
      '--ngProbeLocations',
      server.root_dir,
    }

    opts.cmd = cmd
    opts.on_new_config = function(new_config)
      new_config.cmd = cmd
    end
  end
```
and it was fine, but when I realized this was still there and removed it (leaving only the plugin's config) it stopped working.

PR #282, opened 2021-11-24
Added some clarity to eslint readme

saltsucker:
Feel free to let me know what you would like changed to match your syntax/style.

I know the *NOTE* is obvious to most, but for noobies like me, it may not be. :)

### DIFF
--- a/lua/nvim-lsp-installer/servers/eslint/README.md
+++ b/lua/nvim-lsp-installer/servers/eslint/README.md
@@ -58,4 +58,5 @@ lsp_installer.on_server_ready(function (server)
     server:setup(opts)
 end)
 ```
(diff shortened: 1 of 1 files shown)


lua/nvim-lsp-installer/servers/eslint/README.md
```diff
@@ -56,3 +58,4 @@ lsp_installer.on_server_ready(function (server)
     server:setup(opts)
 end)
 ```
+*NOTE*: You will have to install `eslint` package locally to your `node_modules` or you will get `Unable to find ESLint library.`
```
williamboman, line 61: It should be defaulting to a global installation of eslint if it fails to resolve a local installation, maybe this can be rephrased to

```suggestion

*NOTE*: You will have to install the [`eslint` package](https://www.npmjs.com/package/eslint) either locally or globally for the server to run successfully.
```

williamboman (approved):

PR #284, opened 2021-11-25
added explicit expand-archive command

boris-kuz:
There's two packages that provide the `Expand-Archive`, (see https://github.com/PowerShell/Microsoft.PowerShell.Archive/issues/52). It can happen that the wrong `Expand-Archive` is chosen, resulting in an error. This fixes that.


williamboman (approved):
Cool! LGTM

I assume this is compatible with all versions of the builtin powershell?

williamboman:
> I assume this is compatible with all versions of the builtin powershell?

Seems like there's [docs for it](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/?view=powershell-5.1) across all versions, 🤷

Issue #124, opened 2021-10-04
Support solang for solidity


ChristianChiarulli:
You can grab the binaries here: https://github.com/hyperledger-labs/solang/releases

williamboman:
Does https://github.com/williamboman/nvim-lsp-installer/pull/125 (`add-solang` branch) work?

You don't happen to have a Windows machine you could try it on as well 😅?

PR #296, opened 2021-11-29
solidity_ls language server

edag94:
Adding solidity language server using published version of [juanblanco/vscode-solidity](https://github.com/juanfranblanco/vscode-solidity) vscode extension, which supports autocomplete and other features currently not supported by the solang LSP this repo contains


edag94:
Seems on the compiler side, this is still a work in progress: https://github.com/ethereum/solidity/issues/7763. Seems that for now, solang and vscode-solidity are the closest to something viable working.

For #124 , I was unsuccessful with setting up solang server to handle import maps (`solang --importmap`) in neovim. The solang compiler seems to handle it successfully when not in `--language-server` mode, but when used in neovim as a language server, for @OpenZeppelin packages for example, the language server cannot find it. With import maps not working correctly, and the solang server not providing autocompletions anyway, I don't think it currently provides a great LSP experience like in other languages.

Inspired by: https://github.com/neovim/nvim-lspconfig/issues/1102, I tried porting the language server from vscode-solidity to be used in nvim by publishing the server from the extension and adding an installer, and was successful with this PR. It seems the compiler errors that show up in vscode do not show up when using nvim, but autocompletions are supported, which is what I was looking for the most, and will be using. 

I would like to add my config for the LSP to the repo with this PR so others can have access to it.

ChristianChiarulli:
nice @edag94 , I tried getting that to work like a month ago but was unsuccessful.

williamboman (commented):
Cool! Does this server exist in lspconfig? I can't seem to find a server config for `solidity_vscode` there

edag94:
Let me add it there. I didn't get a chance to figure out how the two repos fit together exactly.

williamboman:
> Let me add it there. I didn't get a chance to figure out how the two repos fit together exactly.

Yeah so all nvim-lsp-installer does is it provides installations for a subset of the server configurations provided by lspconfig. We used to have custom server configs that didn't exist in lspconfig before, but they've all been merged upstream by now. The name of a server in this plugin needs to correspond with a server configuration provided by lspconfig

lua/nvim-lsp-installer/servers/solidity_vscode/init.lua
```diff
@@ -0,0 +1,15 @@
+local server = require "nvim-lsp-installer.server"
+local npm = require "nvim-lsp-installer.installers.npm"
+
+return function(name, root_dir)
+    return server.Server:new {
+        name = name,
+        root_dir = root_dir,
+        languages = { "solidity" },
+        homepage = "https://github.com/edag94/vscode-solidity",
+        installer = npm.packages { "solidity-language-server" },
```
williamboman, line 10: Who's behind this npm package?

I wonder if we instead should just download the VSCode `.vsix` file from their github releases and unpack it instead? Otherwise, I think it might be worthwhile trying to get this package included in https://github.com/hrsh7th/vscode-langservers-extracted. I suggested the same for the [Spectral LSP](https://github.com/luizcorreia/spectral-language-server/issues/1#issuecomment-981547790), which is another standalone npm package that just wraps a VSCode LSP.

williamboman: @hrsh7th 👋 What are your thoughts on how to deal with all the VSCode LSPs repackaged as npm packages? Do you think we should try to centralize all of them in hrsh7th/vscode-langservers-extracted, or would it make sense to have individual packages maintained by separate individuals? If we were to add more servers to hrsh7th/vscode-langservers-extracted, I think we might have to think about package size and perhaps release a separate npm package per VSCode LSP

edag94: @williamboman I'm the package publisher.

I forked the [vscode-solidity repo](https://github.com/juanfranblanco/vscode-solidity) and changed the executable to be the server rather than the extension, so that it would fit the npm.executable input (`cmd = { npm.executable(root_dir, "solidity-language-server"), "--stdio" }`), then published it. I also reached out to the owner of the vscode-solidity repo to let him know, and make him a maintainer of the npm package.

christianparpart: I am not really sure this overhead (of using the vscode extension) is really worth it. Of course, it is definitely better than nothing, and we already agreed with their author to work together, such that vscode-solidity will be utilizing solc in LSP mode once released, but for nvim I think in the long (!) run does not make too much sense once solc does provide this functionality.

I am writing this here purely as FYI. :-)

edag94: @christianparpart what is the timeframe for the language server built into solc? I agree that this is the ideal, I was just looking to provide the community with the most viable temp option until the language server is released, with the plan of deprecating it then.

edag94:
@williamboman Add the language server config here: https://github.com/neovim/nvim-lspconfig/pull/1552. Will update here again once that's merged

edag94:
Solidity_ls has been merged: https://github.com/neovim/nvim-lspconfig/pull/1552
Will work on fixing the conflicts and fixing the name to match the one in nvim-lspconfig

edag94:
Updated naming and fixed conflicts

williamboman (approved):

lua/nvim-lsp-installer/servers/solidity_vscode/init.lua
christianparpart: @edag94 sorry for the late response. IRL is tight lately. So the initial solc LSP release is imminent (this week (tm)). We start with diagnostics only though, we have way more features implemented already, but want to roll out each in their own cycle for maximum stability. Also: you don't need to defend yourself. I'm actually really glad there is interest in this subject at all. This shows the high need of such a solution and that we should address it in a way that is as close as possible to the upstream language and semantics, which can be best achieved when utilizing the already existing native compiler implementation. :)

christianparpart: p.s.: I'm using actually NeoVIM (with coc.nvim) to develop the native LSP ;-)

Issue #292, opened 2021-11-28
Arduino Language Server


thetic:
**Which server would you like to request to be added?**
Server name: arduino_language_server
Server homepage (e.g., GitHub page): https://github.com/arduino/arduino-language-server

**Which languages does this server target?**
Languages: arduino

I have a rudimentary implementation here that probably doesn't work outside of Linux and misses some edge cases:

```lua
local arduino_config = function()
    local server_name = 'arduino_language_server'

    local server = require('nvim-lsp-installer.server')
    local root_dir = server.get_server_root_path(server_name)
    local go = require('nvim-lsp-installer.installers.go')
    local installers = require('nvim-lsp-installer.installers')
    local shell = require('nvim-lsp-installer.installers.shell')

    return server.Server:new({
        name = server_name,
        root_dir = root_dir,
        homepage = 'https://github.com/arduino/arduino-language-server',
        installer = installers.pipe({
            go.packages({
                'github.com/arduino/arduino-language-server',
            }),
            shell.bash(
                go.executable(root_dir, 'arduino-language-server')
                    .. 'config init --overwrite'
            ),
        }),
        default_options = {
            cmd = {
                go.executable(root_dir, 'arduino-language-server'),
                '-cli-config',
                '~/.arduino15/arduino-cli.yaml',
            },
        },
    })
end

require('nvim-lsp-installer.servers').register(arduino_config())
```

williamboman:
Hello! I've actually been looking into this and have a branch ready for it. What's blocking it currently is that the cmd to start the server seems to rely on user-specific configs, such as the `-cli-config` option. Few questions:

- What is the `-cli-config` option? Is there some common path that this could default to, or is it generally always custom to each user? Is it a project-specific file, or a system-wide?
- Do you know if the `-cli` and `-clangd` options are needed? If yes, we could make sure to install `arduino-cli` and `clangd` as well.

thetic:
>     * What does the `arduino-language-server config init --overwrite` command do?

In my case, I just wanted to ensure that a config existed. The server won't run without one.

>     * What is the `-cli-config` option? Is there some common path that this could default to, or is it generally always custom to each user? Is it a project-specific file, or a system-wide?

This is the default location. It appears to be a global config:

```yaml
board_manager:
  additional_urls: []
daemon:
  port: "50051"
directories:
  data: /home/chad/.arduino15
  downloads: /home/chad/.arduino15/staging
  user: /home/chad/Arduino
library:
[...]

PR #298, opened 2021-11-30
Add arduino-language-server supprot

thetic:
Address #292 

This is far from finished.

- [ ] Handle `arduino-cli.exe` (vs `arduino-cli`) on Windows
- [ ] Prevent go installer from using the `arduino-cli` version when installing `arduino-language-server`.


lua/nvim-lsp-installer/servers/arduino_language_server/init.lua
```diff
@@ -0,0 +1,75 @@
...
+                    )
+                ),
+                when(
+                    platform.is_win,
+                    coalesce(
+                        when(platform.arch == "x64", "arduino-cli_%s_Windows_64bit.zip"),
+                        when(platform.arch == "x86", "arduino-cli_%s_Windows_32bit.zip")
+                    )
+                )
+            ):format(version)
```
williamboman, line 40: Calling `:format` on this expression might cause nil pointer errors, for example if you're on mac but on arm64. This function can return `nil`, in which case the installer will interpret it as an error caused by the current system not being supported. What about:

```lua
local target_file = coalesce(...)
return target_file and target_file:format(version)
```

lua/nvim-lsp-installer/servers/arduino_language_server/init.lua
```diff
@@ -0,0 +1,75 @@
...
+        name = name,
+        root_dir = root_dir,
+        homepage = "https://github.com/arduino/arduino-language-server",
+        languages = { "arduino" },
+        installer = {
+            arduino_cli_installer,
+            installers.on {
+                unix = shell.sh(path.concat{ root_dir, "arduino-cli" } .. " config init --dest-file arduino-cli.yaml --overwrite"),
+                win = shell.cmd(path.concat{ root_dir, "arduino-cli.exe" } .. " config init --dest-file arduino-cli.yaml --overwrite")
+            },
```
williamboman, line 62:
```suggestion
            ---@type ServerInstallerFunction
            function (_, callback, ctx)
                process.spawn(path.concat { root_dir, "arduino-cli" }, {
                    args = { "config", "init", "--dest-file", "arduino-cli.yaml", "--overwrite" },
                    cwd = ctx.install_dir,
                    stdio_sink = ctx.stdio_sink,
                }, callback)
            end,
```

williamboman (commented):
I wonder if we should install each component in a separate subdirectory, to avoid them interfering with one another? I don't think there's a nice way of doing so currently, let me think of how such an installer API could look like...

thetic:
I opened an issue on the LSP to mimic `arduino-cli`'s config file deduction: https://github.com/arduino/arduino-language-server/issues/90

williamboman (approved):
Nice work 🎯! I think this looks good now. This is installing 3 different components atm, and because there are no builtin APIs to help with this, they all install to the same directory. I mention it in another comment earlier, but I'd like to make that possible before merging this. Will look into it asap and push any necessary changes

williamboman:
@thetic I added a commit that makes sure to install the CLI and LSP to separate directories. Could you see if things still work for you? I also removed the arch check for Mac, because it seems like it starts just fine on my arm Mac (although I couldn't really get any diagnostics or go to definition working in a sample project I found on GitHub, but I might be doing something wrong).

williamboman (approved):

thetic:
This isn't working for me. I wonder if clangd is indeed required.

williamboman:
> This isn't working for me. I wonder if clangd is indeed required.

Did my commits break things, or was it not working before either?

thetic:
I'm investigating now.

thetic:
Disregard. I'm running into limitations of the actual server it seems. Installation is working as expected.

williamboman:
> Disregard. I'm running into limitations of the actual server it seems. Installation is working as expected.

Yeah the server seems to not really do anything? Let's merge this for now, I'd be happy to bundle clangd at a later point if it turns out to make sense to do so

PR #302, opened 2021-12-03
fix(angularls): also use new_root_dir for probe locations

### LINKED ISSUE #300: Angularls does not work without adding to ngProbeLocations path
**Describe the bug**
Out of the box angularls does not work for me without changing `root_dir` to `root_dir .. "/node_modules/@angular/language-server"` for the `--ngProbeLocations` param. 

**Expected behavior**
A month ago I used angularls with this plugin and it worked without issue out of the box.
Today I was attempting to set up my configuration on a new mac and ran into an issue where the `ngProbeLocations` path could not be resolved by ngserver. I then re-installed angularls on my old mac and this issue came up there too. For now I have just fixed it in my local handler but i'm not sure if this is a change we would want to make in the plugins config itself.

**System info (please complete the following information):**
 - OS: [e.g. macOS 11.6]
 - Neovim version: v0.6.0

williamboman:
Fixes #300.

### DIFF
--- a/lua/nvim-lsp-installer/servers/angularls/init.lua
+++ b/lua/nvim-lsp-installer/servers/angularls/init.lua
@@ -2,15 +2,6 @@ local server = require "nvim-lsp-installer.server"
 local npm = require "nvim-lsp-installer.installers.npm"
 
 return function(name, root_dir)
-    local cmd = {
-        npm.executable(root_dir, "ngserver"),
-        "--stdio",
-        "--tsProbeLocations",
-        root_dir,
-        "--ngProbeLocations",
-        root_dir,
-    }
-
     return server.Server:new {
         name = name,
         root_dir = root_dir,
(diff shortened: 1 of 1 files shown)


mjlbach:
Why not fix things upstream in lspconfig so everyone benefits?

williamboman:
Not sure if this could be added to lspconfig. We install a local typescript installation alongside the LSP server. If this were to be added to lspconfig it'd probably end up looking something like:

```lua
function get_cmd(probe_dir)
  -- Detect if typescript is installed globally, if it is, determine its file path
  local global_tsserver = a.wait(job(...))
  -- Determine where ngserver is installed globally, use it for "--ngProbeLocation" because the language service is a direct dependency to ngserver
  local global_ngserver = a.wait(job(...))

  return { "ngserver", ... }
end
```

This is how VSCode bundles their angular "extension" by the way. It ships with typescript installed.

williamboman:
And so what was added in this PR is how lspconfig currently works. Prior to this PR we'd only use the bundled typescript and angular dependencies

mjlbach:
I tried to do the auto-detection here: https://github.com/neovim/nvim-lspconfig/pull/1462

It's possible to test the project local packages to see if a project local typescript is installed

williamboman:
> It's possible to test the project local packages to see if a project local typescript is installed

This is actually how it's configured now. You can apparently provide more than one location to probe, so we provide (in this order, because that's how VSCode does it 🤷 ):

1) the lsp installer location where typescript is installed
2) the workspace directory

Issue #304, opened 2021-12-04
Pyright not getting installed if setup overridden by lspconfig customizations


smacz42:
Similar to[An iissue Originally posted by @HagaiHargil](https://github.com/williamboman/nvim-lsp-installer/issues/276#issuecomment-977734451), after installing Pyright I get the issue:

```
Error detected while processing FileType Autocommands for "python":                                                                                                                                                                           
E5108: Error executing lua /usr/share/nvim/runtime/lua/vim/lsp/rpc.lua:384: start `pyright-langserver` failed: ENOENT: no such file or directory                                                                                              
```

I have this in my config where I have lspconfig installed:

```
local lsp = require('lspconfig')

lsp.pyright.setup{
    settings = {
        python = {
            analysis = {
                extraPath = {'.'},
                autoImportCompletions = true,
            }
        }
    }
}
```

This is how I _believe_ it is suggested to override configs. However, when I comment this out I do NOT get the error.

I believe that the nvim-lsp-installer sets the path by setting the `cmd` attribute using variables in this plugin's scope, that may not be available in the scope for lspconfig.

A quick fix here is to add right in the top level, the full path to the command to run:

```
cmd = {os.getenv("HOME") .. '/.local/share/nvim/lsp_servers/python/node_modules/.bin/pyright-langserver', '--stdio'},
```

However, I know that this plugin tries to set that up. Seems a waste to have to override it if there is any customiation to the config.

williamboman:
Hello! I just wrote an answer to a similar issue here https://github.com/williamboman/nvim-lsp-installer/issues/305#issuecomment-986030260, hopefully it makes sense. tl;dr is that the `cmd` is only automatically injected if you set up the server via nvim-lsp-installer.

williamboman:
And here's an example snippet in the wiki https://github.com/williamboman/nvim-lsp-installer/wiki/Advanced-Configuration#overriding-the-default-lsp-server-options

smacz42:
Ok, that works.

Just so you know where I was coming from (and in case anyone else gets tripped up by this), I had installed [Plug Config](https://github.com/ouuan/vim-plug-config), whose paradigm is to put all associated configs settings in files separated by plugin.

So, when I went to override the lspconfig, I went to put it in a separate file from the installer setup. However, since these two plugins are so closely intertwined, it makes sense in this situation to break that paradigm and overlap those plugin configs.

Issue #301, opened 2021-12-03
Make readme more noob friendly


juligreen:
I spent quite some time trying to understand how to get the lsp keybinds described in the nvim-lspconfig's README until I read https://github.com/williamboman/nvim-lsp-installer/issues/244 .

Could this project's README include that you should remove the configuration from nvim-lspconfig?
Also it would be nice if it could also include the keybinds from the original readme.
Kind of like this:

```
local lsp_installer = require("nvim-lsp-installer")

-- Register a handler that will be called for all installed servers.
-- Alternatively, you may also register handlers on specific server instances instead (see example below).
lsp_installer.on_server_ready(function(server)
    local opts = {}

    -- (optional) Customize the options passed to the server
    -- if server.name == "tsserver" then
    --     opts.root_dir = function() ... end
    -- end

    -- This setup() function is exactly the same as lspconfig's setup function.
    -- Refer to https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md
    server:setup(opts)
  
  -- Add keybinds
  local function buf_set_keymap(...) vim.api.nvim_buf_set_keymap(bufnr, ...) end
  local function buf_set_option(...) vim.api.nvim_buf_set_option(bufnr, ...) end

  -- Enable completion triggered by <c-x><c-o>
  buf_set_option('omnifunc', 'v:lua.vim.lsp.omnifunc')

  -- Mappings.
  local opts = { noremap=true, silent=true }

  -- See `:help vim.lsp.*` for documentation on any of the below functions
  buf_set_keymap('n', 'gD', '<cmd>lua vim.lsp.buf.declaration()<CR>', opts)
  buf_set_keymap('n', 'gd', '<cmd>lua vim.lsp.buf.definition()<CR>', opts)
  buf_set_keymap('n', 'K', '<cmd>lua vim.lsp.buf.hover()<CR>', opts)
  buf_set_keymap('n', 'gi', '<cmd>lua vim.lsp.buf.implementation()<CR>', opts)
  buf_set_keymap('n', '<C-k>', '<cmd>lua vim.lsp.buf.signature_help()<CR>', opts)
  buf_set_keymap('n', '<space>wa', '<cmd>lua vim.lsp.buf.add_workspace_folder()<CR>', opts)
  buf_set_keymap('n', '<space>wr', '<cmd>lua vim.lsp.buf.remove_workspace_folder()<CR>', opts)
[...]

PR #306, opened 2021-12-04
docs: clarify how servers are installed and the recommended setup

### LINKED ISSUE #305: Adding newly installed server to PATH automatically
I have been trying to understand how to use this for the past 2 days. How does this plugin work? Do I have to add the language server to my `PATH` every time I install a new server or should it handle everything itself and I am doing something wrong?
After installing `clangd` with `:LspInstall clangd` and opening a c++ file, I am getting this error:

<code>
Spawning language server with cmd: `clangd` failed. The language server is either not installed, missing from PATH, or not executable.</code>

However this does not occur after manually adding clangd to PATH. Also, `:LspInstallLog` show no errors.

I am leaving this issue unlabeled because I don't even know if this is expected behavior. I am using the default config.

williamboman:
Resolves #305, #304, #301.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,18 @@ Plug 'williamboman/nvim-lsp-installer'
 
 ### Setup
 
+nvim-lsp-installer installs server executables in a local directory that doesn't exist on PATH. In order for the neovim
+LSP client to be able to locate these executables, the full path to the executable needs to be provided when setting up
+a server. In lspconfig, this is provided via the `cmd` property in the table provided to the `.setup()` function, for
+example (`lspconfig.sumneko_lua.setup { cmd = { "/path/to/lua-server" } }`).
+
+The recommended way of setting up your installed servers is to do it directly through nvim-lsp-installer. By doing so,
+nvim-lsp-installer will make sure to inject any necessary properties before calling lspconfig's setup function for you.
+You may find a minimal example below. To see how you can override the default settings for a server, refer to the
+[Wiki](overriding-default-settings).
+
+[overriding-default-settings]: https://github.com/williamboman/nvim-lsp-installer/wiki/Advanced-Configuration#overriding-the-default-lsp-server-options
+
 ```lua
 local lsp_installer = require("nvim-lsp-installer")
 
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ example (`lspconfig.sumneko_lua.setup { cmd = { "/path/to/lua-server" } }`).
 The recommended way of setting up your installed servers is to do it directly through nvim-lsp-installer. By doing so,
 nvim-lsp-installer will make sure to inject any necessary properties before calling lspconfig's setup function for you.
 You may find a minimal example below. To see how you can override the default settings for a server, refer to the
-[Wiki](overriding-default-settings).
+[Wiki][overriding-default-settings].
 
 [overriding-default-settings]: https://github.com/williamboman/nvim-lsp-installer/wiki/Advanced-Configuration#overriding-the-default-lsp-server-options
 
--- a/doc/nvim-lsp-installer.txt
+++ b/doc/nvim-lsp-installer.txt
@@ -42,6 +42,22 @@ https://github.com/williamboman/nvim-lsp-installer/blob/main/CUSTOM_SERVERS.md.
 ==============================================================================
 QUICK START                                   *nvim-lsp-installer-quickstart*
 
+nvim-lsp-installer installs server executables in a local directory that
+doesn't exist on PATH. In order for the neovim LSP client to be able to locate
+these executables, the full path to the executable needs to be provided when
+setting up a server. In lspconfig, this is provided via the `cmd` property in
+the table provided to the `.setup()` function, for example: >
+
+    lspconfig.sumneko_lua.setup { cmd = { "/path/to/lua-server" } }
+<
+
+The recommended way of setting up your installed servers is to do it directly
+through nvim-lsp-installer. By doing so, nvim-lsp-installer will make sure to
+inject any necessary properties before calling lspconfig's setup function for
+you. You may find a minimal example below. To see how you can override the
+default settings for a server, refer to the Wiki:
+https://github.com/williamboman/nvim-lsp-installer/wiki/Advanced-Configuration#overriding-the-default-lsp-server-options.
+
 To view the UI for nvim-lsp-installer, run: >
 
     :LspInstallInfo


williamboman:
@juligreen @smacz42 @VanillaBase1lb 👋, does adding this make things clearer?

mjlbach:
I don't understand why it is necessary to hijack the setup process from lspconfig. With the changes I made to accommodate lsp-installer, you should be able to globally or per_server inject cmd_env with path. If you need additional callbacks exposed in lspconfig to make this easier, that's fine (and doable), but I see more and more reports of users getting confused because there are now two defacto ways of setting the same underlying autocommand, one of which is wrapping the other.

VanillaBase1lb:
If the `nvim-lsp-installer` setup is going to hijack `lspconfig`'s setup function, I think it is worth mentioning in the readme that this setup functions should/can be called after settings keybindings from `lspconfig`'s default config in https://github.com/neovim/nvim-lspconfig#keybindings-and-completion. Otherwise the keybindings will not work, even if defined in init file.

williamboman:
> I don't understand why it is necessary to hijack the setup process from lspconfig.

There are a couple of reasons for this:

There are some additional, I/O bound, overhead introduced by this plugin where we have to identify which servers are installed, and for those installed we need to `require` some extra Lua modules. These operations currently get scheduled as to not impair perceived startup times. This causes some friction with how setting servers up via lspconfig alone works, where you have to imperatively setup each individual server somewhere in your init script.

The above could be alleviated if lspconfig fully supported setting up servers after the init script has executed. Basically, if the following worked, it'd remove one reason to "wrap" lspconfig:

```lua
-- Somewhere in your init script
vim.defer_fn(function ()
  lspconfig.sumneko_lua.setup { }
end, 5000)
```

```sh
$ nvim file.lua
# After waiting 5 seconds - I'd expect the Lua server to be attached to the buffer
```

Last time I investigated this, lspconfig assumes servers are only ever set up during init, and so it doesn't attach to any open buffers. 

Then there's the requirement for nvim-lsp-installer to actually inject properties to the server config, most commonly just being `cmd`/`cmd_env` but there are cases where additional properties are needed. The wrapped `server:setup()` method does this for you.

These the main motivations behind the wrapped `setup()` function.

----

I'd like avoid users calling `lspconfig.server.setup()` on servers that aren't installed. What exists in ones init script doesn't really govern which servers are actually installed and available, but only which servers one would like to enable. I also want to make it easier to just start using a new LSP server without having to add actual code to your config. If you just provide a callback to nvim-lsp-installer that is responsible of setting up a server, you can very easily get started with new servers using lspconfig's default settings.

These are the main motivations behind the `lsp_installer.on_server_ready(function () ... end)` API. It allows deferring the setting up of servers til we know they're actually installed and ready to start.

----

> With the changes I made to accommodate lsp-installer, you should be able to globally or per_server inject cmd_env with path. 

Hmm do you mean by modifying the tables in the module cache? Like

```lua
local configs = require("lspconfig.configs")
configs.sumneko_lua.cmd = { ... }
```

I've kinda wanted to avoid doing this, at least through this particular mechanism (directly modifying the table in module cache).

williamboman:
To add to the above, I think it'd be neat if the lspconfig APIs separated the action of providing ones custom settings for a server and actually enabling the server. Something like:

```lua
local lspconfig = require("lspconfig")

-- 1. Providing ones settings first, and then 2. calling the `.enable()` function on the table returned by `.setup()`
lspconfig.sumneko_lua.setup({
  custom_settings = {
     i_have_these_preferences = "very nice"
  }
}).enable()

-- If using another 3rd party plugin like nvim-lsp-installer, we only provide the settings and delegate the "enable" action to it
lspconfig.sumneko_lua.setup({
  custom_settings = {
     i_have_these_preferences = "very nice"
  }
})

-- Then, nvim-lsp-installer can just do `lspconfig.sumneko_lua.enable()` when it detects the server is installed
```

mjlbach:
> There are some additional, I/O bound, overhead introduced by this plugin where we have to identify which servers are installed, and for those installed we need to require some extra Lua modules. 

I don't understand why this can't be solved by storing some state in vim.stdpath('data')/cache nor why you would require extra lua modules conditionally per server.

> This causes some friction with how setting servers up via lspconfig alone works, where you have to imperatively setup each individual server somewhere in your init script.

The startup time impact should be minimal, especially for users using impatient. The most "expensive" part of lspconfig is calling vim.lsp which brings in the built-in client code, which IMO is inevitable.

> The above could be alleviated if lspconfig fully supported setting up servers after the init script has executed. Basically, if the following worked, it'd remove one reason to "wrap" lspconfig:

Lspconfig does support setting up servers after the init script. You just need to manually trigger the filetype autocommand to get the server to attach to the current buffer.

> I've kinda wanted to avoid doing this, at least through this particular mechanism (directly modifying the table in module cache).

I don't see why you wouldn't. There's also a global on_setup hook you can use.
https://github.com/mjlbach/nvim-lspconfig/blob/8436a197cc76f33fc2dc26145272c66cf5ee193d/lua/lspconfig/configs.lua#L53-L55

> To add to the above, I think it'd be neat if the lspconfig APIs separated the action of providing ones custom settings for a server and actually enabling the server. 

It already does, by allowing you to modify default_config by accessing the configs table.

williamboman:
> I don't understand why this can't be solved by storing some state in vim.stdpath('data')/cache nor why you would require extra lua modules conditionally per server.

It probably would alleviate some things, but it would carry some other complexity that I'd like to avoid if possible. The act of detecting which servers is installed is extremely quick (it just scans a directory), the bottleneck is predominantly all the `require` calls. As for extra Lua modules, what I was referring to was the individual server modules that contain the default options, [like this one](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/hls/init.lua#L32-L37).

> Lspconfig does support setting up servers after the init script. You just need to manually trigger the filetype autocommand to get the server to attach to the current buffer.

Yeah this is one of the reasons why nvim-lsp-installer has its own `server:setup()` method which makes sure to attach any open buffers to the server.

> It already does, by allowing you to modify default_config by accessing the configs table.

I think it'd be neat with a dedicated API for it, something like https://github.com/williamboman/nvim-lsp-installer/pull/306#issuecomment-986063395. I think having users, or 3rd party plugins, reach into and mutate the tables themselves brings too tight coupling, not to mention complexity with writing to nested table structures:

```lua
local configs = require("lspconfig.configs")
-- Can I write directly to the full path, will the default option table structure always allow this?
configs.eslint.default_options.settings.format.enable = true
-- or do I have to extend it
vim.tbl_deep_extend("force", configs.eslint.default_options, {
  settings = {
     format = { enable = true }
  }
})
```

From the use case of this plugin, the order of execution also becomes important because we'd then also have to ensure that this plugin is able to execute code that modifies these default settings prior to the user calling `.setup{}` in their init script. But maybe I'm missing something?

PR #311, opened 2021-12-05
fix(powershell_es): don't limit to windows

lizalc:
The [`PowerShellEditorServices`](https://github.com/PowerShell/PowerShellEditorServices) are cross-platform. This allows installing on more than just Windows.

Currently using this on Apple silicon.

### DIFF
--- a/lua/nvim-lsp-installer/servers/powershell_es/init.lua
+++ b/lua/nvim-lsp-installer/servers/powershell_es/init.lua
@@ -2,7 +2,6 @@ local server = require "nvim-lsp-installer.server"
 local path = require "nvim-lsp-installer.path"
 local std = require "nvim-lsp-installer.installers.std"
 local context = require "nvim-lsp-installer.installers.context"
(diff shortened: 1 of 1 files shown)


williamboman (approved):
Cool! Are there any other runtime requirements to run it? In some places we don't allow installing a server if a certain runtime requirement is not available (see usages of `std.ensure_executables` for example)

lizalc:
It depends on PowerShell 7+ ([ref](https://github.com/PowerShell/PowerShellEditorServices#1-install-powershell-7)) so the `pwsh` executable needs to be available. This is true even on Windows where `powershell.exe` is available. I've updated my changes.

Message when `pwsh` is unavailable:

```
Pending servers (1)
  ◍ powershell_es (failed)
      pwsh was not found in path. Refer to https://github.com/PowerShell/PowerShell#get-powershell for instructions.
```

Let me know if there is anything else I missed!

williamboman (approved):

PR #315, opened 2021-12-06
fix(quick_lint_js): account for the fact that the archive has different structures across platforms

williamboman:


williamboman:
Ping @strager you might be interested in this. I expected the different .zip & .tar.gz archives to share the same directory structure, but it seems like the .tar.gz come wrapped with a `quick-lint-js/` root directory, whereas the .zip does not

Issue #317, opened 2021-12-07
Add more info about pylsp installer


Fymyte:
**Realy hard to find how to add packages for plugins using `pip3` installer**
After digging for 2 days in lsp logs and codes for pylsp I finally understood how to add pylsp_mypy and pyls_isort plugins.
lsp installed with `pip3` installer seems to create a venv in the lsp_installer directory, thus not finding packages in the user system.
To install those plugins I had to source the created venv and use pip in it.

**Possible solutions**
In the first time, I think it would be great to add better installation instructions for lsp that uses pip3 installer and explicitly says that they use a virtual env, and how to add packages for them.
An even better solution would be to add the opportunity for the user to customize an existing server, to add additional packages to install in the venv. I don't know how this would be implemented tho, It's just an idea.

**Describe alternatives you've considered**
For now I had to create an entirely new server and register it in my config, just to add those missing packages automatically.
```lua
local function ensure_lsp_installed(servers)
  for _, name in pairs(servers) do
    local server_is_found, server = lsp_installer.get_server(name)
    if server_is_found then
      if not server:is_installed() then
        utils.log('warn', 'Server ' .. name .. ' not installed. Installing...')
        server:install()
      end
    end
  end
end

local custom_pylsp_name = 'pylsp-full'
local lspconfig_configs = require('lspconfig.configs')
local lspconfig_util = require('lspconfig.util')
local pip3 = require('nvim-lsp-installer.installers.pip3')
local lsp_installer_servers = require('nvim-lsp-installer.servers')
local lsp_installer_server = require('nvim-lsp-installer.server')
local root_dir = lsp_installer_server.get_server_root_path(custom_pylsp_name)
lspconfig_configs[custom_pylsp_name] = {
  default_config = {
    cmd = { 'pylsp' },
    filetypes = { 'python' },
    root_dir = function(fname)
      local root_files = {
        'pyproject.toml',
        'setup.py',
        'setup.cfg',
        'requirements.txt',
        'Pipfile',
      }
      return lspconfig_util.root_pattern(unpack(root_files))(fname) or lspconfig_util.find_git_ancestor(fname)
    end,
    single_file_support = true,
  },
}
local custom_pylsp_server = lsp_installer_server.Server:new( {
  name = custom_pylsp_name,
  root_dir = root_dir,
[...]

PR #318, opened 2021-12-07
add :PylspInstall command for managing 3rd party pylsp plugins

Fymyte:
#317
Try adding better documentation to install `pylsp` plugins using `nvim-lsp-installer`
I don't think it completely fixes the issue as it only add documentation, so it should not close the issue

### DIFF
--- a/lua/nvim-lsp-installer/servers/pylsp/README.md
+++ b/lua/nvim-lsp-installer/servers/pylsp/README.md
@@ -1,0 +1,21 @@
+# Pylsp
+
+## Installing pylsp plugins
+pylsp is installed in a virtual environment. If you want to install additional
+python packages and make them available to pylsp, you have to install them in
+the virtual environment.
+
+To find the installation of pylsp:
+```vim
+:lua print(vim.fn.stdpath('data') .. '/lsp_servers/pylsp')
+```
+cd into this directory and source the virtual env:
+```sh
+source venv/bin/activate
+```
+
+then you can install plugins as you would globally.
+ex of `pylsp_mypy`:
+```sh
+pip install pylsp_mypy
+```
--- a/lua/nvim-lsp-installer/servers/pylsp/README.md
+++ b/lua/nvim-lsp-installer/servers/pylsp/README.md
@@ -3,10 +3,7 @@
 ## Installing pylsp plugins
 Pylsp has [third party plugins](https://github.com/python-lsp/python-lsp-server#3rd-party-plugins) which are not installed by default.
 
-If you want to install them using the same installation of pylsp as nvim-lsp-installer, you may run
+In order for these plugins to work with the `pylsp` server managed by this plugin, they need to be installed in the same [virtual environment](https://docs.python.org/3/library/venv.html) as `pylsp`. For these reasons, there's a convenient `:PylspInstall <packages>` command that does this for you, for example:
 ```vim
-:PylspInstall <list of python packages>
+:PylspInstall pyls-flake8 pylsp-mypy pyls-isort
 ```
-
-PS:
-pylsp is installed in a python venv located in `stdpath('data')/lsp_servers/pylsp`
--- a/lua/nvim-lsp-installer/servers/pylsp/README.md
+++ b/lua/nvim-lsp-installer/servers/pylsp/README.md
@@ -1,21 +1,12 @@
 # Pylsp
 
 ## Installing pylsp plugins
-pylsp is installed in a virtual environment. If you want to install additional
-python packages and make them available to pylsp, you have to install them in
-the virtual environment.
+Pylsp has [third party plugins](https://github.com/python-lsp/python-lsp-server#3rd-party-plugins) which are not installed by default.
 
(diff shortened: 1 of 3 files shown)


Fymyte:
The command `PylspInstall` is added and works well with packages pip knows.
Stderr is printed at the end of the command. I don't know if it is something we want, or just print them as log in the `lsp_installer.log` file ?

lua/nvim-lsp-installer/servers/pylsp/README.md
```diff
@@ -0,0 +1,12 @@
+# Pylsp
+
+## Installing pylsp plugins
+Pylsp has [third party plugins](https://github.com/python-lsp/python-lsp-server#3rd-party-plugins) which are not installed by default.
+
+If you want to install them using the same installation of pylsp as nvim-lsp-installer, you may run
+```vim
+:PylspInstall <list of python packages>
```
williamboman, line 8:
```suggestion
In order for these plugins to work with the `pylsp` server managed by this plugin, they need to be installed in the same [virtual environment](https://docs.python.org/3/library/venv.html) as `pylsp`. For these reasons, there's a convenient `:PylspInstall <packages>` command that does this for you, for example:

```vim
:PylspInstall pyls-flake8 pylsp-mypy pyls-isort
```

lua/nvim-lsp-installer/servers/pylsp/README.md
```diff
@@ -0,0 +1,12 @@
...
+```
+
+PS:
+pylsp is installed in a python venv located in `stdpath('data')/lsp_servers/pylsp`
```
williamboman, line 12: Do you think is needed for any particular reason? I think it might be a bit overkill (also, since the installation directory is customizable this isn't entirely true)

```suggestion
```

lua/nvim-lsp-installer/servers/pylsp/init.lua
```diff
@@ -10,6 +12,32 @@ return function(name, root_dir)
         installer = pip3.packages { "python-lsp-server[all]" },
         default_options = {
             cmd = { pip3.executable(root_dir, "pylsp") },
+            commands = {
+                PylspInstall = {
+                    function(...)
+                        -- `nargs+` requires at least one argument -> no empty table
+                        local plugins = { ... }
+                        local plugins_str = table.concat(plugins, ", ")
+                        notify(string.format("Installing %q...", plugins_str))
```
williamboman, line 21:
```suggestion
                        notify(("Installing %q..."):format(plugins_str))
```

lua/nvim-lsp-installer/servers/pylsp/init.lua
```diff
@@ -10,6 +12,32 @@ return function(name, root_dir)
...
+                        process.spawn(
+                            pip3.executable(root_dir, "pip"),
+                            {
+                                args = vim.list_extend({ "install", "-U", "--disable-pip-version-check" }, plugins),
+                                stdio_sink = process.simple_sink(),
+                            },
+                            vim.schedule_wrap(function(success)
+                                if success then
+                                    notify(string.format("Successfully installed %q", plugins_str))
```
williamboman, line 30:
```suggestion
                                    notify(("Successfully installed %q"):format(plugins_str))
```

lua/nvim-lsp-installer/servers/pylsp/init.lua
```diff
@@ -10,6 +12,32 @@ return function(name, root_dir)
...
+                                else
+                                    notify("Failed to install requested plugins.", vim.log.levels.ERROR)
+                                end
+                            end)
+                        )
+                    end,
+                    description = "Install listed modules in the env of Pylsp.",
```
williamboman, line 37:
```suggestion
                    description = "Installs the provided packages in the same venv as pylsp.",
```

williamboman (commented):
Awesome! Some suggestions for the docs, and also some minor stylistic changes in the Lua, otherwise LGTM!

lua/nvim-lsp-installer/servers/pylsp/README.md
Fymyte: At least for mypy, we might want to know from where to execute some commands:
[from mypy docs](https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports) you have to run the binary to install some missing type stubs.

williamboman: Ah I see. It seems like you can also install the `types-PyYAML` pip package also, so that is at least easy via `:PylspInstall` :+1:.

If it turns out to be use cases around accessing a shell with the venv activated, we can consider another command that opens a new terminal window with the venv activated for you, like:

```vim
command! PylspVenv exe 'tabnew | term' | startinsert | call feedkeys("source ~/.local/share/nvim/lsp_servers/pylsp/venv/bin/activate^M")
```

williamboman (approved):
great stuff

Fymyte:
Thanks for guiding me.
It has made a great first experience for open source projects

PR #330, opened 2021-12-12
fix: zls-0.2 uses a different archive path

rslabbert:
See #329 for bug report: zls-0.2 change the archive format.

I'm new to the codebase so I'm not sure if this is the best way to implement the fix, but it's tested as working locally for me. Happy to make quick changes or close in favour of a different fix!


williamboman (approved):
LGTM!

PR #336, opened 2021-12-16
Add sourcekit-lsp for Swift

jesse-c:
sourcekit-lsp isn't attempted to be installed. It verifies if it is already installed. This is the approached used for [Dart](https://github.com/williamboman/nvim-lsp-installer/pull/267).

I'm having some issues running `./scripts/autogen_metadata.sh` on my machine.


williamboman:
> I'm having some issues running `./scripts/autogen_metadata.sh` on my machine.

Ah, this is because the name needs to match the [lspconfig server name](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/sourcekit.lua). I pushed a new commit that renames it :+1:

williamboman (approved):
Great!

jesse-c:
Ah, that was silly of me. Sorry about that! Thank you for fixing it up.

PR #343, opened 2021-12-18
fix(health): ignore the patch number in capturing go version

Alienover:
### Reason
The pattern `go(%d+)%.(%d+)%.(%d+)` can **only** be matched when the `go version` contains the patch number (e.g. go1.17.3).
In my case, the go version I installed is exactly `1.17`.
image

Before:
image

After:
image

Test result for the `go version` with/without patch number
```lua
string.find("go version go1.17 darwin/amd64", "go(%d+)%.(%d+)")
-- 12      17      1       17

string.find("go version go1.17.3 darwin/amd64", "go(%d+)%.(%d+)")
-- 12      17      1       17
```


williamboman (approved):
Ah thanks! I blame copilot (I guess my comment instructing it didn't help 🤣)

PR #345, opened 2021-12-19
add zk

thecontinium:
Added zk a lsp server for markdown managed through the zk tool.


lua/nvim-lsp-installer/servers/zk/init.lua
```diff
@@ -0,0 +1,18 @@
+local server = require "nvim-lsp-installer.server"
+local std = require "nvim-lsp-installer.installers.std"
+
+return function(name, root_dir)
+    return server.Server:new {
+        name = name,
+        root_dir = root_dir,
+        homepage = "https://https://github.com/mickael-menu/zk",
+        languages = { "markdown" },
+        installer = std.ensure_executables {
```
williamboman, line 10: Any reason you decided not to actually install zk here? It seems like it's pretty straight forward, they even provide prebuilt binaries?

This approach of only checking whether the executables exist is only applied in rare cases where it doesn't make much sense to install the LSP locally. Dart and Sourcekit (Swift) are the only two so far - reason being they both ship the LSP server as part of their SDK; so all we do is provide instructions on how to install these separately. This doesn't seem to be the case here?

lua/nvim-lsp-installer/servers/zk/init.lua
```diff
@@ -0,0 +1,18 @@
+local server = require "nvim-lsp-installer.server"
+local std = require "nvim-lsp-installer.installers.std"
+
+return function(name, root_dir)
+    return server.Server:new {
+        name = name,
+        root_dir = root_dir,
+        homepage = "https://https://github.com/mickael-menu/zk",
```
williamboman, line 8:
```suggestion
        homepage = "https://github.com/mickael-menu/zk",
```

williamboman (commented):
Cool! I think we should actually make sure to install the server as well, unless there's a reason not to which I'm not seeing. Let's also add it to the README!

thecontinium:
Sure - Can you point me to a good example that installs the latest tag with a non constant/versioned name ? Perhaps using a regex ? Cheers.

williamboman:
> Can you point me to a good example that installs the latest tag with a non constant/versioned name ?

Sure! The [denols](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/denols/init.lua) server is probably very similar. The call to `context.use_github_release_file(repo, release_file)` will call the GitHub API to fetch the latest release for the given `repo`, and then using the resolved release version it'll produce the URL to the `release_file` asset (in this case it'd be either `zk-%s-linux-amd64.tar.gz`, `zk-%s-linux-arm64.tar.gz`, [and so on](https://github.com/mickael-menu/zk/releases/tag/v0.9.0)). For Windows, this function can simply return `nil`, which means no release file is available for the current OS.

I see they also do `.tar.gz` for Linux releases, and `.zip` for macOS, so that needs to be taken into account in the next installer step, which accesses the context via `context.capture()` (see `std.untargz_remote()` and `std.unzip_remote()`)

williamboman (approved):
Ty @thecontinium! I like the way you used `coalesce` & `when` as top-level expressions - elsewhere I've usually wrapped them with a function and explicitly had to handle nil values

Issue #357, opened 2021-12-24
Error while installing phpactor


shurizzle:
**Describe the bug**
phpactor gives an error when I try to install it

**Expected behavior**
No errors.

**Screenshots**
![immagine](https://user-images.githubusercontent.com/203655/147372465-95faf2c1-a74f-48b1-939d-08a9b30f5531.png)

**System info (please complete the following information):**

-   OS: Ubuntu 20.04 on WSL2
-   Neovim version: v0.7.0-dev
-  nvim-lsp-installer: git main

Checkhealth output (`:checkhealth nvim-lsp-installer`):

```

nvim-lsp-installer: require("nvim-lsp-installer.health").check()
========================================================================
## nvim-lsp-installer report
  - WARNING: **Go**: not available
  - WARNING: **java**: not available
  - OK: **sh**: `Ok`
  - OK: **tar**: `tar (GNU tar) 1.30`
  - OK: **gzip**: `gzip 1.10`
  - OK: **curl**: `curl 7.68.0 (x86_64-pc-linux-gnu) libcurl/7.68.0 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3`
  - OK: **wget**: `GNU Wget 1.20.3 built on linux-gnu.`
  - OK: **python3**: `Python 3.8.10`
  - OK: **node**: `v16.13.1`
  - OK: **Ruby**: `ruby 2.7.0p0 (2019-12-25 revision 647ee6f091) [x86_64-linux-gnu]`
  - OK: **bash**: `GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)`
  - OK: **PHP**: `PHP 7.4.3 (cli) (built: Nov 25 2021 23:16:22) ( NTS )`
  - OK: **RubyGem**: `3.1.2`
  - OK: **Composer**: `Composer 1.10.1 2020-03-13 20:34:27`
  - OK: **npm**: `8.1.2`
  - OK: **pip3**: `pip 21.1.1 from /home/shura/.local/lib/python3.8/site-packages/pip (python 3.8)`
```

Please also include any other system information you think is relevant!

**Install log**

```
[...]

PR #358, opened 2021-12-24
Fixing fs.rmrf and std.git_clone

shurizzle:
Referring to issue #357 

Now fs.rmrf doesn't flood the log file with fake errors.
std.git_clone accepts opts (the second argument) as optional without giving errors

### DIFF
--- a/lua/nvim-lsp-installer/fs.lua
+++ b/lua/nvim-lsp-installer/fs.lua
@@ -21,9 +21,11 @@ end
 function M.rmrf(path)
     log.debug("fs: rmrf", path)
     assert_ownership(path)
-    if vim.fn.delete(path, "rf") ~= 0 then
-        log.debug "fs: rmrf failed"
-        error(("rmrf: Could not remove directory %q."):format(path))
+    if vim.fn.exists(path) then
+        if vim.fn.delete(path, "rf") ~= 0 then
+            log.debug "fs: rmrf failed"
+            error(("rmrf: Could not remove directory %q."):format(path))
+        end
     end
 end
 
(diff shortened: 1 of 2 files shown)


lua/nvim-lsp-installer/installers/std.lua
```diff
@@ -239,7 +239,7 @@ function M.git_clone(repo_url, opts)
         opts = vim.tbl_deep_extend("force", {
             directory = ".",
             recursive = false,
-        }, opts)
+        }, opts or {})
```
williamboman, line 242: This should be the culprit!

lua/nvim-lsp-installer/fs.lua
```diff
@@ -21,9 +21,11 @@ end
 function M.rmrf(path)
     log.debug("fs: rmrf", path)
     assert_ownership(path)
-    if vim.fn.delete(path, "rf") ~= 0 then
-        log.debug "fs: rmrf failed"
-        error(("rmrf: Could not remove directory %q."):format(path))
+    if vim.fn.exists(path) then
```
williamboman, line 24: Although this may be the sensible thing to do, it breaks the current contract of this function - where it errors if the provided path does not exist. I think it'd be best to not introduce this diff

williamboman (commented):
Ah thanks! I hope to get https://github.com/williamboman/nvim-lspconfig-test integrated in a better way as it would've caught this.

shurizzle:
> Ah thanks! I hope to get https://github.com/williamboman/nvim-lspconfig-test integrated in a better way as it would've caught this.

I apprecciate a lot your work, keep it on :)

williamboman:
Superseded by b77e86ef40bee93aa1b8652fd3ed6de6a7cb12eb. Thanks @shurizzle!

PR #363, opened 2021-12-26
feat(ui): display interactive server settings schema table

williamboman:
https://user-images.githubusercontent.com/6705160/147408227-85439ddf-b194-43e9-adf0-245352e8c337.mov


williamboman:
@kylo252 wdyt?

kylo252:
This actually looks incredible already! 

Any reason to not keep using json? Here's a production-ready script used by [nlsp-settings](https://github.com/tamago324/nlsp-settings.nvim) for this (although it could definitely benefit from using `vim.json` instead): https://github.com/tamago324/nlsp-settings.nvim/blob/053495bfc97c9c61bebf740acbbcaeb1ebc0704c/scripts/gen_schemas.lua

williamboman:
> Any reason to not keep using json? Here's a production-ready script used by [nlsp-settings](https://github.com/tamago324/nlsp-settings.nvim) for this (although it could definitely benefit from using `vim.json` instead): https://github.com/tamago324/nlsp-settings.nvim/blob/053495bfc97c9c61bebf740acbbcaeb1ebc0704c/scripts/gen_schemas.lua

Do you mean storing the schema in JSON? Just to give some more background - I figured there are 3 alternative approaches:

1. Fetch and store the schema (if available) in the server installation directory at the point of installation, which could then be easily accessed and parsed via `vim.json.decode`
1. Fetch the schema ad-hoc when interacting with the UI
1. Store the schemas directly in the plugin (either as Lua or JSON)

I decided to go with 3. because it was least complex, and it will only ever render static, known, JSON. To answer your question - I went with Lua to leverage its builtin module resolution. I'm actually not sure how one would load JSON files in the repo w/o resorting to hacks like using the `debug` package to produce the paths.

Also some more reasoning behind this addition: it's not meant to be a definitive source for server settings, but only a cursory overview of which settings are available (and more importantly signal to users that settings actually is a thing).

PR #364, opened 2021-12-26
fix(arduino_language_server): install clangd, also add docs on how to work with FQBN

williamboman:


williamboman:
Hey @thetic - what do you think about this? I had to do some research to understand what's what (no Arduino experience) and I figured this would be the best approach. Apparently it seems like the language server is somewhat tightly coupled with their new Arduino IDE (based on Theia) - there could be a case for a dedicated neovim plugin to deal with the rough edges 👀

williamboman:
With this I'm actually getting diagnostics and all the other LSP goodies, which I wasn't before 🤷

thetic:
This looks like a significant improvement. Nice work and thank you. I only wish we could extract this from arduino config file located with the same strategy as the Arduino CLI instead of hardcoding paths into nvim's config.

https://arduino.github.io/arduino-cli/0.19/configuration/#locations

thetic:
I've tested this out in my setup and it is working very well so far.

PR #371, opened 2021-12-29
add solc

### LINKED ISSUE #368: [New Server Request]: Integrate solc 0.8.11's LSP server
**Which server would you like to request to be added?**

-   Server name: `solc` LSP
-   Server homepage (e.g., GitHub page): https://github.com/ethereum/solidity/releases/tag/v0.8.11

**Which languages does this server target?**

-   Languages: `solidity`

**How is this server distributed?**

-   Is the server distributed through a standardized channel (such as release files, npm, pip, etc.)? (leave empty if unknown): there are [js bindings](https://github.com/ethereum/solc-js) but that introduces overhead. If possible it'd be preferable to pull a system-dependent release from their github releases channel.
-   Is this server added in [lspconfig](https://github.com/neovim/nvim-lspconfig)?: Not for `solc`

williamboman:
Resolves #368.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ lsp_installer.settings({
 | SQL                                 | `sqls`                    |
 | Solidity                            | `solang`                  |
 | Solidity                            | `solc`                    |
-| Solidity                            | `solidity_ls`             |
+| Solidity (VSCode)                   | `solidity_ls`             |
 | Sorbet                              | `sorbet`                  |
 | Sphinx                              | `esbonio`                 |
 | Stylelint                           | `stylelint_lsp`           |
--- a/README.md
+++ b/README.md
@@ -212,8 +212,9 @@ lsp_installer.settings({
 | Rust [(wiki)][rust_analyzer]        | `rust_analyzer`           |
 | SQL                                 | `sqlls`                   |
 | SQL                                 | `sqls`                    |
-| Solang Solidity                     | `solang`                  |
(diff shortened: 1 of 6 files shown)


williamboman:
https://github.com/neovim/nvim-lspconfig/pull/1623

chriseth:
In my opinion, the github release page mechanism is seriously flawed since it does not provide a history of changes.
I would recommend to use `https://binaries.soliditylang.org` instead, which is (mirrored from) a proper git repository and also allows you to get a list of releases. The structure of the repository is explained here: https://docs.soliditylang.org/en/v0.8.11/installing-solidity.html#static-binaries

williamboman:
Thanks for the input @chriseth! I took a quick look, shouldn't be any problems. I guess one drawback would be that installing older versions might be a bit cumbersome from an end user perspective, as you'd have to provide the full `v0.3.6+commit.988fe5e5` component (`:LspInstall solc@v0.3.6+commit.988fe5e5`), whereas through GitHub releases the tag name alone would suffice.

Great documentation btw :+1:! If you don't mind some unsolicited feedback; one thing that wasn't immediately clear to me was how to access the files via https://binaries.soliditylang.org/ (the landing page links to the docs, which links to the landing page). Took me a minute or two to piece things together. I think hosting a file explorer (like https://c.quick-lint-js.com/) would help with self discovery!

Issue #194, opened 2021-10-24
[Feature] Add a `LspUpdateAll` command


5c077m4n:
Can you please add a `LspUpdateAll` command that updates all of the installed servers and restarts them so they'll replace the old ones?

williamboman:
A little sneak peak

https://user-images.githubusercontent.com/6705160/141648667-94218eb8-95ad-46c9-884f-3e65b393d92c.mov

williamboman:
Update: The complexity in implementing this (in a robust and undiscriminating<sup>1</sup> manner) is that there's virtually no information stored about installed servers - other than the fact that it's "installed" (its directory exists on the filesystem). The only cases when some sort of metadata stored is when a server is installed through a package manager such as `npm`, as npm will capture some metadata in `package.json`. After some experimentation I think I've settled on an approach, where an installation receipt will be saved on the fs (as JSON), to be used for persisting metadata that otherwise is only available in-memory at the point of installation.

Also, I've merged #353 which gives easy access to update all installed servers through the UI window (default keybind is `U`).

<sup>1 - what I mean by this is that servers that are installed through entirely custom methods (such as building something from source) should not be excluded</sup>

williamboman:
See https://github.com/williamboman/nvim-lsp-installer/pull/395 for an initial implementation of checking for outdated servers. It's currently only tied to the UI (no command yet, will add later, will have to think about rate limiting and whatnot first).

Please give it a try if you'd like (branch `outdated-servers`). Note that it will only work with your _currently installed_ servers if you've recently updated the plugin version (it requires at least https://github.com/williamboman/nvim-lsp-installer/pull/379) _and_ recently updated/installed a server. If you install or update a server it will of course be able to check its versions going forward.

lucax88x:
:LspInstallInfo works wonderfully.

Issue #248, opened 2021-11-08
Option to provide GitHub API token?


rosenhouse:
Hi, I'm trying this out, but my machine is behind a shared IP and we often hit API rate limiting from GitHub.

e.g. on my machine
```
$ curl https://api.github.com/repos/clangd/clangd/releases/latest
{"message":"API rate limit exceeded for <MY_IP_ADDRESS>. (But here's the good news: Authenticated requests get a higher rate limit. Check out t
he documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}
```

so when I go to `:LspInstall something` I see 
```
   ▏   ▏   ▏   ▏   ▏   ▏   ▏   ▏   ▏   ▏     nvim-lsp-installer
   ▏   ▏   ▏   ▏   ▏   ▏   ▏   ▏   ▏   ▏      press ? for help
   ▏   ▏   ▏   ▏   ▏   ▏     https://github.com/williamboman/nvim-lsp-installer

 Installed servers (1)
   ◍ rust_analyzer

 Pending servers (1)
   ◍ clangd (failed)
   ▏   Fetching latest release version from GitHub API...
   ▏   Failed to fetch url "https://api.github.com/repos/clangd/clangd/releases/latest".
   ▏   curl: (22) The requested URL returned error: 403
```

Is there a way I can provide a token as an environment variable or other option, so that lsp-installer can use it when doing requests to GitHub?  I've seen projects support environment variables like `GITHUB_TOKEN` or `GITHUB_API_TOKEN`.

Thank you!

kylo252:
How about using [gh](https://github.com/cli/cli) by default when it's available? It'll handle the authentication for you without having to pass around any API keys to the environment, and it can even perform `jq` queries if needed.

We can also make it opt-in
```lua
settings = {
	---
	use_gh_cli = false,
}
```

kylo252:
> I really want to be conservative with dependencies. Personally, had I seen a plugin like this requiring `gh` I'd nope out immediately.

@williamboman, just a reminder that it can be completely optional. I was thinking that it should help solve any rate-limiting issues with github, and not to actually completely replace anything. So basically check if it's there and then use `fetcher()` otherwise. 

It would also mean that you just tell people to use it if they're having trouble with the default `curl` or whatever the default `fetcher()` happens to be. That would be easier and more maintainable than trying to create a cross-platform OAuth solution.

[...]

PR #395, opened 2022-01-05
feat(ui): display outdated servers

williamboman:
https://user-images.githubusercontent.com/6705160/148303390-6449551e-c820-4c36-816e-2b1da69ec4ae.mov

Ping #194.


williamboman:
@kylo252 👋  any opinions?

lua/nvim-lsp-installer/jobs/outdated-servers/git.lua
```diff
@@ -0,0 +1,42 @@
+local process = require "nvim-lsp-installer.process"
+local CheckResult = require "nvim-lsp-installer.jobs.outdated-servers.check-result"
+
+---@param server Server
+---@param source InstallReceiptSource
+---@param on_check_complete fun(result: CheckResult)
+return function(server, source, on_check_complete)
+    process.spawn("git", {
+        -- We assume git installation track the remote HEAD branch
+        args = { "fetch", "origin", "HEAD" },
```
kylo252, line 10: I'm not sure that this is reliable, since you might be starting with a shallow clone, so you might want to pass a `--force` flag. An alternative that I've found working relatively well for me is using [git ls-remote](https://git-scm.com/docs/git-ls-remote.html), which removes the hassle of dealing with `--unshallow` or deepening the history of a giant repo. 

Here's an example
https://github.com/LunarVim/LunarVim/blob/b10e18a213741dcd4ce1e7ba63accaf10719dbec/lua/lvim/utils/git.lua#L146-L154

lua/nvim-lsp-installer/jobs/outdated-servers/github_release_file.lua
```diff
@@ -0,0 +1,29 @@
+local github = require "nvim-lsp-installer.core.clients.github"
+local CheckResult = require "nvim-lsp-installer.jobs.outdated-servers.check-result"
+
+---@param server Server
+---@param source InstallReceiptSource
+---@param on_result fun(result: CheckResult)
+return function(server, source, on_result)
+    github.fetch_latest_release(
```
kylo252, line 8: it's probably time to add `gh` integration? 😄

kylo252 (commented):
I really like the idea! 

~~triggering the update-check on the info panel directly will mean that you clearly intend that panel to be used once per-session. So checking filetypes, settings or anything else might end up being cumbersome. I do however see the benefit, and moving it into each server block might reduce visibility.~~ I didn't notice it was supposed to be triggered by a keymap and it's not working for me.

williamboman:
> triggering the update-check on the info panel directly will mean that you clearly intend that panel to be used once per-session.

No not really. There are two primary reasons it's only done automatically once per session:
1) Low effort attempt to avoid triggering rate limits
2) To signal to users that the feature exist - to make it easily discoverable without having to read a README. This is for example the same reason why the server is automatically expanded when it's installed.

The panel can be used as many times as one wants to.

lua/nvim-lsp-installer/jobs/outdated-servers/github_release_file.lua
williamboman: Hehe I don't mind writing these HTTP client calls manually. This is however starting to feel a whole lot like early days Node.js dev. I wouldn't be surprised if core primitives for async/await, HTTP, fs, path find its way into neovim core (and then lambda expressions in LuaJIT, somehow) :D.

Also, I want to keep amount of dependencies minimal. Once (and if) the clienty stuff in lspconfig makes it to core and all what remains are server configs, I would actually even consider vendoring/bundling those inside this plugin entirely.

williamboman:
> I didn't notice it was supposed to be triggered by a keymap and it's not working for me.

Hmm, does `:nmap <buffer> c` and `:nmap <buffer> C` produce anything?

kylo252:
It's working correctly on this machine at least, maybe I had something messed before. What I suspected from the demo was true,  the check-version does fire automatically. 

> triggering the update-check on the info panel directly will mean that you clearly intend that panel to be used once per-session. So checking filetypes, settings or anything else might end up being cumbersome.

Also what about potential errors or slow-downs? maybe the automatic check should be opt-in?

lua/nvim-lsp-installer/jobs/outdated-servers/git.lua
kylo252: Yeah it does work, hopefully there are no gotchas related to the installed version of git.

williamboman: > Yeah it does work, hopefully there are no gotchas related to the installed version of git.

I have a solution in store if this is the case:
608p65

williamboman:
> Also what about potential errors or slow-downs? maybe the automatic check should be opt-in?

Errors are actually not visible in the UI currently, they should show up in the logs though. As for slowdowns I'm not too worried. I've not measured it but the processes should all mostly be IO-bound. Right now it's hard-capped at 4 concurrent processes. I thought about making it possible to disable, I think it makes sense tbh. But I'd rather make it opt-out, unless there's a strong case to make it opt-in that I'm not seeing?

kylo252:
>  I've not measured it but the processes should all mostly be IO-bound. Also, it only runs when you open the UI window. Right now it's hard-capped at 4 concurrent processes.

The main realistic improvement I can think of is a timeout

**curl**
```console
$ curl --help all | rg time
     --connect-timeout <fractional seconds> Maximum time allowed for connection
     --expect100-timeout <seconds> How long to wait for 100-continue
     --happy-eyeballs-timeout-ms <milliseconds> Time for IPv6 before trying IPv4
     --keepalive-time <seconds> Interval time for keepalive probes
 -m, --max-time <fractional seconds> Maximum time allowed for transfer
 -R, --remote-time        Set the remote file's time on the local output
     --retry-delay <seconds> Wait time between retries
     --retry-max-time <seconds> Retry only within this period
 -y, --speed-time <seconds> Trigger 'speed-limit' abort after this time
 -z, --time-cond <time>   Transfer based on a time condition
     --trace-time         Add time stamps to trace/verbose output
```
**npm**
```console
$ npm config list -l | rg time
fetch-retry-maxtimeout = 60000
fetch-retry-mintimeout = 10000
fetch-timeout = 300000
```

**powershell**

```console
$ Get-Help Invoke-WebRequest -Parameter TimeoutSec

-TimeoutSec <int>

    Required?                    false
    Position?                    Named
    Accept pipeline input?       false
    Parameter set name           (All)
    Aliases                      None
    Dynamic?                     false
    Accept wildcard characters?  false
```

> I thought about making it possible to disable, I think it makes sense tbh. But I'd rather make it opt-out, unless there's a strong case to make it opt-in that I'm not seeing?

The main one I'm thinking of is the rate limiting, e.g. github's API

williamboman:
> The main realistic improvement I can think of is a timeout

Ah yeah good point. I think implementing that in the Lua layer by sending a SIGKILL to the process handle after a certain amount of time would be easier (and completely tool agnostic). Question is what is a reasonable timeout? 30s?

> The main one I'm thinking of is the rate limiting, e.g. github's API

Yeah I ran into this multiple times during dev and had to constantly switch VPN connection. I wonder if the best course of action is to keep it as-is and try to solve this holistically in #248.

kylo252:
> Ah yeah good point. I think implementing that in the Lua layer by sending a SIGKILL to the process handle after a certain amount of time would be easier (and completely tool agnostic). Question is what is a reasonable timeout? 30s?

I would've said something similar, but according to Microsoft: 
> > A Domain Name System (DNS) query can take up to 15 seconds to return or time out. If your request contains a host name that requires resolution, and you set TimeoutSec to a value greater than zero, but less than 15 seconds, it can take 15 seconds or more before a WebException is thrown, and your request times out.

So yeah.. maybe we use the same one as `winget`? https://github.com/microsoft/winget-cli/blob/cfd8b296f3339f6ef6b714af8d4f9959a78fc9c1/schemas/JSON/settings/settings.schema.0.2.json#L110-L116

> Yeah I ran into this multiple times during dev and had to constantly switch VPN connection. I wonder if the best course of action is to keep it as-is and try to solve this holistically in #248.

I'm still in the opinion that it's easier to tell people to install `gh` and be done with it, since it'll handle the authentication for you without having to pass around any "naked" API keys to the environment, and it can even perform `jq` queries. 

Something to consider, is security. since it's hard(-er) to avoid dealing with that API key without loading it into memory, which would expose it to all the plugins using it. Sure, you can tell the fetcher to read it from a file, but now you have to worry about making sure to _not_ log it. Am I overthinking this, and no ones care? 😅

williamboman:
> I'm still in the opinion that it's easier to tell people to install `gh` and be done with it, since it'll handle the authentication for you without having to pass around any "naked" API keys to the environment, and it can even perform `jq` queries.

I really want to be conservative with dependencies. Personally, had I seen a plugin like this requiring `gh` I'd nope out immediately.

> Something to consider, is security. since it's hard(-er) to avoid dealing with that API key without loading it into memory, which would expose it to all the plugins using it. Sure, you can tell the fetcher to read it from a file, but now you have to worry about making sure to _not_ log it. Am I overthinking this, and no ones care? 😅

It's all valid points. I haven't given it much thought yet, but what I'm thinking so far is to attempt using the `gh` utility for all GitHub API calls, and if it fails fall back to the existing `PowerShell (Windows only) -> wget -> curl` chain.

williamboman:
Let's merge this. I think there's a high risk that GitHub API rate limit will be a problem (albeit a benign one). More incentives to start looking into #248 if anything!

Issue #326, opened 2021-12-11
New server request: nimlsp


Yabanahano:
**Which server would you like to request to be added?**
Server name: `nimlsp`
Server homepage (e.g., GitHub page): https://github.com/PMunch/nimlsp

**Which languages does this server target?**
Languages: `nim`

williamboman:
Cool! At first glance it looks like compiling the project would be the best path, as `nimble` seems to be the only other option. I'm not familiar with nimble myself, but it seems like it only really installs to a "global" location, which is not particularly compatible with this plugin. I also found https://github.com/pragmagic/vscode-nim, do you know if they're the same? I assume they are since the lspconfig entry [links to both](https://github.com/pragmagic/vscode-nim)

zetashift:
> I also found https://github.com/pragmagic/vscode-nim, do you know if they're the same? 

They kind of aren't! Both nimlsp and the vscode extensions rely on `nimsuggest`, a tool that comes with Nim so provide better tooling support.

I agree that the best path will be compiling the project. 
One thing to note is this from nimlsp's README:

```
When compiling and using nimlsp it needs to have Nim's sources available in order to work. 
With Nim installed through choosenim these should already be on your system 
and nimlsp should be able to find and use them automatically
```

Most Nim users will have choosenim installed, but not all!

williamboman:
Added in #406

PR #396, opened 2022-01-06
Added nimls

BliznyukNM:
For now you have to install nim by yourself.

### DIFF
--- a/lua/nvim-lsp-installer/servers/nimls/init.lua
+++ b/lua/nvim-lsp-installer/servers/nimls/init.lua
@@ -1,0 +1,21 @@
+local server = require "nvim-lsp-installer.server"
+local shell = require "nvim-lsp-installer.installers.shell"
+
+return function(name, root_dir)
+  return server.Server:new {
+    name = name,
+    root_dir = root_dir,
+    homepage = "https://github.com/PMunch/nimlsp",
+    languages = { "nim" },
+    installer = {
+      shell.bash [[ 
+        git clone --depth 1 https://github.com/PMunch/nimlsp.git .
+        nimble build
+      ]]
+    },
+    default_options = {
+      cmd = { root_dir.."/nimlsp" },
+      filetypes = { "nim" },
+    },
+  }
+end
--- a/lua/nvim-lsp-installer/servers/init.lua
+++ b/lua/nvim-lsp-installer/servers/init.lua
@@ -73,6 +73,7 @@ local CORE_SERVERS = Data.set_of {
     "kotlin_language_server",
     "lemminx",
     "ltex",
+    "nimls",
     "ocamlls",
     "omnisharp",
     "phpactor",


lua/nvim-lsp-installer/servers/nimls/init.lua
```diff
@@ -0,0 +1,21 @@
...
+    name = name,
+    root_dir = root_dir,
+    homepage = "https://github.com/PMunch/nimlsp",
+    languages = { "nim" },
+    installer = {
+      shell.bash [[ 
+        git clone --depth 1 https://github.com/PMunch/nimlsp.git .
+        nimble build
+      ]]
+    },
```
williamboman, line 15: To make this OS-agnostic, we should avoid using the `shell` installers altogether (I don't believe they're used anywhere in this plugin anymore, they'll most likely be removed soon).

Something like this ought to do it (I added the `context.receipt()` call, which will make sure that write a receipt with installation information to a JSON file upon completion.

```suggestion
    installer = {
      std.git_clone "https://github.com/PMunch/nimlsp.git",
      function (_, callback, ctx)
        process.spawn("nimble", {
          args = { "build" },
          cwd = ctx.install_dir,
          stdio_sink = ctx.stdio_sink,
        }, callback)
      end,
      context.receipt(function (receipt)
        receipt:with_primary_source(receipt.git_remote "https://github.com/PMunch/nimlsp.git")
      end)
    }
```

lua/nvim-lsp-installer/servers/nimls/init.lua
```diff
@@ -0,0 +1,21 @@
...
+    default_options = {
+      cmd = { root_dir.."/nimlsp" },
+      filetypes = { "nim" },
```
williamboman, line 18: These options will essentially be merged with the lspconfig options, and [they already provide](https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/nimls.lua#L5) a `cmd` and `filetypes` for `nimls`. The only thing we need to do is extend the `PATH` environment, so that the client is able to locate the executable:

```suggestion
      cmd_env = {
          PATH = process.extend_path { root_dir },
      }
```

williamboman (commented):
Cool! They write something about [nimls sources potentially being problematic when building from source](https://github.com/PMunch/nimlsp#compile-nimlsp). I did not really understand what that meant, or whether the `-d:explicitSourcePath=PATH` flag needs to be added.

https://github.com/williamboman/nvim-lsp-installer/issues/326#issuecomment-996065450 mentions that `choosenim` is a widely used tool, but we shouldn't assume that everyone has it installed. I wonder if going with the `nimble install nimlsp` option would be more robust? We just need to find a way to make `nimble` install to a local directory.

williamboman:
Superseded by #406.

PR #399, opened 2022-01-06
update server installation issue template to a form-based template

kylo252:
I was in the middle of updating this file https://github.com/LunarVim/LunarVim/blob/8fd71b50847addb160d58c2045ced8d6f6c6e0ef/tests/minimal_lsp.lua, when it occurred to me that it actually belongs here instead. So I figured I can adapt it a bit better and clean up the issue template to mention it.

Hopefully you haven't _purposefully_ avoided having a such a file before, but if so feel free to close this draft 😅


williamboman (commented):
This looks interesting, it seems like a major improvement! Will look through this asap

.github/ISSUE_TEMPLATE/server_installation_form.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - type: input
+    attributes:
+      label: nvim-lsp-installer version
+      description: commit hash, `git log --pretty=%h -1`
+      placeholder: 284e914
+  - type: input
+    attributes:
+      label: nvim-lspconfig version
+      description: commit hash, `git log --pretty=%h -1`
+      placeholder: 299e6395
```
williamboman, line 39: I think these are a bit overkill. What about a simple checkbox? (I'm not sure if the following is valid, need to find the schema for this file)

```suggestion
  - type: checkbox
    attributes:
      label: I've recently downloaded the latest plugin versions
      description: Make sure you've downloaded the latest plugin versions.
```

.github/ISSUE_TEMPLATE/server_installation_form.yaml
```diff
@@ -0,0 +1,88 @@
...
+    validations:
+      required: true
+  - type: textarea
+    attributes:
+      label: Steps to reproduce
+      description: Steps to reproduce using the minimal debug config.
+      placeholder: |
+        You can download it from here: https://github.com/williamboman/nvim-lsp-installer/blob/main/tests/minimal_debug_init.lua
+        1. `nvim -u ./minimal_debug_init.lua`
+        2. ...
```
williamboman, line 60: I feel like this would be better suited for the other issue type (the one that doesn't apply to the installation of servers). I feel like the process of installing servers is already pretty isolated, regardless of the surrounding environment, so being able to reproduce it using a minimal config isn't really that important, imo?

.github/ISSUE_TEMPLATE/server_installation_form.yaml
```diff
@@ -0,0 +1,88 @@
...
+  - type: textarea
+    id: logs
+    attributes:
+      label: log and support info
+      placeholder: |
+        :LspInstallInfo
+        :LspInstallLog
+        :LspInfo
+        :messages
+        :checkhealth
```
williamboman, line 80: I think `:LspInfo` and `:messages` might not be needed for this issue template, or at least I don't see a need?

```suggestion
        :LspInstallInfo
        :LspInstallLog
        :checkhealth nvim-lsp-installer
```

williamboman (commented):
Awesome 👏! I added some suggestions in the interest of lowering the effort needed to submit an issue. As for the minimal debug, in my experience I feel like it's not particularly useful for debugging the installation of servers per se, but more about debugging issues that occur after the fact. For those kinds of issues there exists another issue template ("general-issue.md"), I think it'd make more sense to incorporate it there instead? We could do this in another PR!

.github/ISSUE_TEMPLATE/server_installation_form.yaml
kylo252: Oh this is for the actual operation of the server, maybe server not launching, not attaching or another runtime issue that could be verified with a quick minimal setup. That's why I had the strict version checks in there.

I'll create a separate one for servers installation if you'd like. I didn't really consider that you had one specifically for the installation part.

williamboman (approved):

PR #403, opened 2022-01-06
feat: integrate gh-cli for GitHub API calls

### LINKED ISSUE #248: Option to provide GitHub API token?

williamboman:
Closes #248.

### DIFF
--- a/lua/nvim-lsp-installer/core/clients/github.lua
+++ b/lua/nvim-lsp-installer/core/clients/github.lua
@@ -19,7 +19,12 @@ function M.fetch_releases(repo, callback)
             return callback("Failed to fetch GitHub releases.", nil)
         end
         callback(nil, vim.json.decode(response))
-    end)
+    end, {
+        custom_fetcher = {
+            cmd = "gh",
+            args = { "api", ("repos/%s/releases"):format(repo) },
+        },
+    })
 end
 
 ---@alias FetchLatestGithubReleaseOpts {tag_name_pattern:string}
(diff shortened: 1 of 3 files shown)


kylo252 (commented):
I have mixed feelings about the location of this implementation. I don't see a reason why `gh api` is treated as a fetcher, but for now it works I guess? You can always extract it into its own module, but the thing I'm worried about is introducing that `custom_fetcher` which YAGNI outside of this use-case, probably.. 🤔 

Why not implement this as a sub function?

williamboman:
Yeah it feels a bit awkward, but I think it's worth the trade-off. The other options I can think of would require a bunch of boilerplate that would enables nested retries

PR #406, opened 2022-01-09
add nimls

williamboman:
Co-authored-by: Nikita Bliznyuk <bliznyuknm@gmail.com>

### DIFF
--- a/lua/nvim-lsp-installer/servers/nimls/init.lua
+++ b/lua/nvim-lsp-installer/servers/nimls/init.lua
@@ -1,0 +1,32 @@
+local server = require "nvim-lsp-installer.server"
+local std = require "nvim-lsp-installer.installers.std"
+local process = require "nvim-lsp-installer.process"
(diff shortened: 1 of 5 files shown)


lua/nvim-lsp-installer/servers/nimls/init.lua
```diff
@@ -0,0 +1,35 @@
...
+
+return function(name, root_dir)
+    return server.Server:new {
+        name = name,
+        root_dir = root_dir,
+        homepage = "https://github.com/PMunch/nimlsp",
+        languages = { "nim" },
+        installer = {
+            std.ensure_executable {
+                { "choosenim", "choosenim is required to install nimlsp." },
```
williamboman, line 15: It seems like choosenim is very strongly recommended. If one installs nim through some other means, you'll also be responsible for downloading nim sources. Choosenim does this for you. This seems like a fine trade-off

Issue #422, opened 2022-01-13
[New Server] Psalm


JSteitz:
**Which server would you like to request to be added?**

-   Server name: psalm
-   Server homepage (e.g., GitHub page): https://psalm.dev/

**Which languages does this server target?**

-   Languages: php

**How is this server distributed?**

-   Is the server distributed through a standardized channel (such as release files, npm, pip, etc.)? (leave empty if unknown): composer
-   Is this server added in [lspconfig](https://github.com/neovim/nvim-lspconfig)?: yes

Sorry for not creating a pull request, I just needed the server just now.
The code would be:
```lua
-- init.lua
local installers = require "nvim-lsp-installer.installers"
local path = require "nvim-lsp-installer.path"
local server = require "nvim-lsp-installer.server"
local composer = require "nvim-lsp-installer.installers.composer"
local std = require "nvim-lsp-installer.installers.std"
local context = require "nvim-lsp-installer.installers.context"
local process = require "nvim-lsp-installer.process"

return function(name, root_dir)
    return server.Server:new {
        name = name,
        root_dir = root_dir,
        homepage = "https://psalm.dev/",
        languages = { "php" },
        installer = installers.when {
            unix = {
                std.git_clone "https://github.com/vimeo/psalm.git",
                composer.install(),
                context.receipt(function(receipt)
                    receipt:with_primary_source(receipt.git_remote "https://github.com/vimeo/psalm.git")
                end),
            },
        },
        default_options = {
            cmd_env = {
                PATH = process.extend_path { path.concat { root_dir, "bin" } },
            },
        },
    }
end
```

JSteitz:
PR created.

PR #426, opened 2022-01-15
add server psalm

JSteitz:
Had now time to make a PR for #422


init.lua
```diff
@@ -0,0 +1,30 @@
...
+        languages = { "php" },
+        installer = installers.when {
+            unix = {
+                std.git_clone "https://github.com/vimeo/psalm.git",
+                composer.install(),
+                context.receipt(function(receipt)
+                    receipt:with_primary_source(receipt.git_remote "https://github.com/vimeo/psalm.git")
+                end),
+            },
+        },
```
williamboman, line 23: I wonder if we should do this the "correct way" and `composer require` to install it. I guess you looked at the only other server that uses composer currently (`phpactor`)? That server was implemented this way because they don't support `composer require` for some weird reason, instead they recommend cloning and `composer install`-ing. Also, `phpactor` explicitly doesn't support Windows because of its usage of signals IIRC, I hope `psalm` supports Windows, in which case we don't need the `installers.when`!

williamboman (commented):
Awesome, thanks 👏 ! I'll look into extending the `composer` module to allow `composer.require`-ing dependencies instead of the `clone -> install` dance.

init.lua
JSteitz: Yes, I looked at phpactor.
Yes, the official docs tell you to install with git pull.
But there is also a composer package maintained, so you can also use composer require for phpactor. (he is also a sponsor and contributor) https://packagist.org/packages/phpactor/phpactor

Do you have an example to install with composer.require? Or do you need to create one first?

williamboman: > But there is also a composer package maintained, so you can also use composer require for phpactor. (he is also a sponsor and contributor) https://packagist.org/packages/phpactor/phpactor

Ah I see! I like to avoid using non-official distribution channels because it's an attack vector waiting to be exploited.

> Do you have an example to install with composer.require? Or do you need to create one first?

We used to have this in the code actually, so I'll just use `git reflog -p -- lua/nvim-lsp-installer/installers/composer.lua` to find it again. However, it will also entail some other things like adding a new installation "receipt" type and then implementing an "outdated version check" that goes through composer and so on - I'll look into this asap so don't worry about it! I'll push this directly to this PR (to your branch) soon, then we can merge :+1:

JSteitz: I'm asking them if the package is official, so we could update it to require: https://github.com/phpactor/phpactor/issues/1338#issue-1104717025

williamboman (approved):

williamboman:
@JSteitz Done! Could you try with the latest commits if everyone works fine for you? I also need to try it on Windows before merge

williamboman:
The outdated check seems to work fine for me (I did `:LspInstall psalm@4.18.0`):

Screenshot 2022-01-15 at 14 37 59

JSteitz:
> 

Yep, Updaing, Uninstalling and Installing worked fine for me.

williamboman:
> Yep, Updaing, Uninstalling and Installing worked fine for me.

Cool. Everything seems (mostly) fine on Windows too, so let's merge here. There's one issue with the cmd definition, I've submitted a PR [upstream in lspconfig](https://github.com/neovim/nvim-lspconfig/pull/1660).

Issue #143, opened 2021-10-10
[Feature] add Julia support


jvaverka:
Execute LanguageServer.jl when opening Julia files.

Jdbeck66:
I’d like to add support for nudging this along as well.

williamboman:
coc-julia seems to do [something interesting](https://github.com/fannheyward/coc-julia/blob/master/src/index.ts#L16-L23).  I've only taken a glance at it, but it seems like they compile the bundled JuliaLS package (which calls into LanguageServer). What it compiles into I have no idea, and I have yet to actually try the coc extension.

solidassassin:
@williamboman it compiles to sysimage (compiled julia session) or binary (self explanatory), I don't use Julia regularly, but I remember that jaot compilation times are notoriously long, so this improves startup time significantly (instant startup as opposed to ~30s). As for package management delegation to Julia I honestly think it would be way more efficient to set up a Github action to compile to releases and then grab it from there (preferably to executable so dependencies would be bundled, which wouldn't oppose the original idea of lsp-installer). That's just my take tho.

williamboman:
Ah I see, thanks for the info! I'm however not going to write and/or host a repository of prebuilt binaries for the Julia LSP myself (I'd barely even consider doing this for servers I personally use every day). This sounds like something that could instead be incorporated in the release pipeline of https://github.com/julia-vscode/LanguageServer.jl itself? I'd also be happy to use a 3rd party vendor if someone wants to maintain and host one!

> preferably to executable so dependencies would be bundled, which wouldn't oppose the original idea of lsp-installer

I'm not sure I'm following this sentence 🙈. Do you mean preferably it'd provide prebuilt binaries (in which dependencies are bundled)?

solidassassin:
> Do you mean preferably it'd provide prebuilt binaries (in which dependencies are bundled)?

Yup. I'll probably try to set up a repo which generates releases with actions requiring as less manual maintenance as possible (since I'm already dealing with university and a full time job :smiley: ), tho I really doubt that you would require active maintenance for something like this (unless breaking changes in `PackageCompiler`, `LanguageServer` or Julia itself are frequent).

williamboman:
We might have a pretty simple solution to this, please see branch `add-julials` (#429) - does it work as expected?

hasundue:
@williamboman It works almost perfectly.
Sure, it's better to have all the necessary things under the plugin's control.

[...]

PR #427, opened 2022-01-15
add julials

hasundue:


hasundue:
Sorry I forgot to update README

williamboman:
Hello! Cool! Have you seen the discussions in #143 btw? Due to how the Julia package manager works, we lose a lot of the control and guarantees that is necessary to ensure the plugin behaves in a reliable manner (this is discussed more in #143). 

I've spent some time today looking into how the VSCdoe extension does this, and it seems like they bundle the LanguageServer and all its dependencies, [like so](https://github.com/julia-vscode/julia-vscode/blob/master/scripts/environments/languageserver/Manifest.toml). The server is started [like so](https://github.com/julia-vscode/julia-vscode/blob/master/src/extension.ts#L229-L240), which runs [main.jl](https://github.com/julia-vscode/julia-vscode/blob/master/scripts/languageserver/main.jl). From this, I've been able to set up a simplified version which seems to be working pretty well actually.

Could you see whether branch `add-julials` (#429) works fine for you? I'm not a Julia dev so I can't really tell. It seems a bit "broken" on projects that only has a `Project.toml` and not a `Manifest.toml` (it wont find dependencies, you need to run `Pkg.instantiate()` and then restart neovim for the LSP to find dependencies).

hasundue:
Oh, sorry, I didn't notice you already worked on julia.
Sure, I'm looking into that.

hasundue:
I just removed my changes which are no longer needed and formatted the script to a "Julia style". https://github.com/williamboman/nvim-lsp-installer/pull/432

williamboman:
> Oh, sorry, I didn't notice you already worked on julia. Sure, I'm looking into that.

No worries! We had actually gotten a bit stuck haha, this PR prompted me to take a 2nd look :)

PR #430, opened 2022-01-16
feat(display): let cursor snap into position

kylo252:
Let the cursor snap into "anchor" position in the info panel.

### DIFF
--- a/lua/nvim-lsp-installer/ui/display.lua
+++ b/lua/nvim-lsp-installer/ui/display.lua
@@ -1,6 +1,7 @@
 local log = require "nvim-lsp-installer.log"
 local process = require "nvim-lsp-installer.process"
 local state = require "nvim-lsp-installer.ui.state"
+local settings = require "nvim-lsp-installer.settings"
 
 local M = {}
 
@@ -223,6 +224,22 @@ function M.delete_win_buf(win_id, bufnr)
     end)
 end
 
+function M.on_cursor_moved(win_id, bufnr)
+    local get_anchor_col = function(line)
+        for _, icon in ipairs(vim.tbl_values(settings.current.ui.icons)) do
+            local pos = vim.fn.stridx(line, icon)
+            if pos ~= -1 then
+                return pos
+            end
+        end
+    end
+    local pos = vim.api.nvim_win_get_cursor(win_id)
+    local line = vim.api.nvim_buf_get_lines(bufnr, pos[1] - 1, pos[1], true)[1]
+    local anchor = get_anchor_col(line)
+    -- dump("win_id: ", win_id, "bufnr: ", bufnr, { pos[1], pos[2] }, anchor)
+    vim.api.nvim_win_set_cursor(win_id, { pos[1], anchor or pos[2] })
+end
+
 function M.new_view_only_win(name)
     local namespace = vim.api.nvim_create_namespace(("lsp_installer_%s"):format(name))
     local bufnr, renderer, mutate_state, get_state, unsubscribe, win_id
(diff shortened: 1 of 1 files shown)


kylo252:
@williamboman, I'm not really sure if you wanted to implement this or not, but I was trying to learn a good way to do this, and I wanted something to test it on 😄 

I wanted to share the basic idea first and see what you think, here's a demo (you can see that I'm not enforcing the movement anywhere else, I have another idea for that!)

<details>
<summary>screencast</summary>

https://user-images.githubusercontent.com/59826753/149663974-fac5241a-3b7a-458f-9038-f9a9611949e1.mp4


</details>

williamboman:
Ah this looks interesting! I think this is a very nice QOL addition :+1:. I wonder if it'd be nicer to leverage the UI data structure (the one that's produced by `status-win.lua` and then translated to buffer-modifying calls in `display.lua`) to express which lines should have this effect applied. Maybe we can use the `CASCADING_STYLE` node - and in the `render_node` return value capture which lines in the buffer that should have this effect. Implementation wise this would probably end up looking very similar to the keybinding mechanism.

So something like

```lua
---@param servers ServerState[]
---@param props ServerGroupProps
local function InstalledServers(servers, props)
    return Ui.Node(Data.list_map(
        ---@param server ServerState
        function(server)
            local is_expanded = props.expanded_server == server.name
            return Ui.CascadingStyleNode({ "CURSOR_SNAP" }, -- This would make the cursor snap to all lines rendered within this UI node
                Ui.HlTextNode {
                    Data.list_not_nil(
                        { settings.current.ui.icons.server_installed, "LspInstallerGreen" },
                        { " " .. server.name, "" },
                        Data.when(server.deprecated, { " deprecated", "LspInstallerOrange" }),
                        Data.when(#server.metadata.outdated_packages > 0, { " new version available", "Comment" })
                    ),
                },
                Ui.Keybind(settings.current.ui.keymaps.toggle_server_expand, "EXPAND_SERVER", { server.name }),
                Ui.Keybind(settings.current.ui.keymaps.update_server, "INSTALL_SERVER", { server.name }),
                Ui.Keybind(settings.current.ui.keymaps.check_server_version, "CHECK_SERVER_VERSION", { server.name }),
                Ui.Keybind(settings.current.ui.keymaps.uninstall_server, "UNINSTALL_SERVER", { server.name }),
                Ui.When(is_expanded, function()
                    return Indent {
                        ServerMetadata(server),
                    }
                end),
            }
        end,
        servers
    ))
end
```

kylo252:
_Sorry for the text wall ahead, but I wanted to explain my thought process a bit 😅_ 

---

That's where my second idea comes in, I want to use `extmarks` as anchors since they offer an extremely powerful API that provides their exact row and col which is exactly what we need, and we probably don't even need to use `stridx()`!

My first experiment was to add the icons as virtual_text instead using `nvim_buf_set_extmark`, and that actually worked, but it turns out that `nvim_buf_add_highlight` already leverages `extmarks`, so yeah.. I didn't need to do any of that extra work 🤦‍♂️ 

Check the result of `nvim_buf_get_extmarks` to see what I mean

```diff
diff --git a/lua/nvim-lsp-installer/ui/display.lua b/lua/nvim-lsp-installer/ui/display.lua
index 889c3ac..55c274d 100644
--- a/lua/nvim-lsp-installer/ui/display.lua
+++ b/lua/nvim-lsp-installer/ui/display.lua
@@ -414,6 +414,7 @@ function M.new_view_only_win(name)
                     vim.api.nvim_win_set_config(opened_win_id, create_popup_window_opts())
                 end
             end
+            log.info(vim.api.nvim_buf_get_extmarks(bufnr, namespace, 0, -1, { details = true }))
         end),
         ---@type fun()
         close = vim.schedule_wrap(function()
```

This gives us a very nice list of anchor points, that we can easily use to snap the cursor 🤖 

If you want to optimize this, we could probably consider having different namespaces just to make the lookup more efficient, although I don't really think it's going to be notiecable whatsoever. It's also simple to filter that table by `hl_group` when using `details = true`.

williamboman:
Or hmm I tried this out now myself and I'm not too sure. I don't see a reason for it to lock the cursor in place on the icon?

williamboman:
@kylo252 Is this something you still want to explore or should we close this? I'm not too keen on restricting/controlling cursor movements like this, but I might be misunderstanding the use case

kylo252:
> @kylo252 Is this something you still want to explore or should we close this? I'm not too keen on restricting/controlling cursor movements like this, but I might be misunderstanding the use case

It's more or less an accessibility/stylistic feature imo. I figured that it's clearer how certain keymaps would behave when the cursor is always snapped to the name of the server. Either way, it's pretty subjective, so feel free to close it if you don't feel it's that relevant and/or applicable.

williamboman:
I like the idea of fixing the cursor to the beginning of the line content (`^`), I think it helps with the discoverability of these items actually being expandable and whatnot. But I'm not too sure about restricting further cursor movements, I like the ability of being to put the cursor above the server names, for example to `:LspInstall <C-r>w` or yanking the server name.

williamboman:
I'll close this for now as I'm not too keen on the current implementation (with the above reasons). I'm happy to reopen if you want to see this happen

PR #432, opened 2022-01-17
Format the script

hasundue:
- Format the script to a "Julia" style
- Remove some changes I made which are no longer needed


williamboman (approved):
Ty!!

PR #436, opened 2022-01-19
fix(go): correct installer for versions of go > 1.11

reynn:
depending on level of support this may need to go further to see what version of Go the user has installed but this was able to finally install `gopls` on my local machine.

### DIFF
--- a/lua/nvim-lsp-installer/installers/go.lua
+++ b/lua/nvim-lsp-installer/installers/go.lua
@@ -32,7 +32,7 @@ function M.packages(packages)
                 pkgs[1] = ("%s@%s"):format(pkgs[1], ctx.requested_server_version)
             end
 
-            c.run("go", vim.list_extend({ "get", "-v" }, pkgs))
+            c.run("go", vim.list_extend({ "install", "-v" }, pkgs))
             c.run("go", { "clean", "-modcache" })
 
             c.spawn(callback)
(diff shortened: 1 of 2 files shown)


smgt:
Since you update the go installer to use `go install` instead of `go get` and `go install` requires you to specify a version to work, other servers will need to have their package name updated also. Example https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/jsonnet_ls/init.lua#L10

smgt:
I created a PR with fixes for all servers that use `go.package`. https://github.com/williamboman/nvim-lsp-installer/pull/438

williamboman:
Thanks! I think there will be some additional changes needed (among other things what @smgt) mentioned. I've been taking a look at this for a while actually but I'm struggling a bit with figuring out the best way of keeping a "manifest" file for these go binaries (think, package.json for npm). I've tried making use of `go.mod` but to no avail.. I don't think this in particular would be a blocker for making this change though, just wanted to bring it up.

To add to what @smgt said, the `@latest` versions shouldn't be hardcoded in the `go.packages` calls. This would make installing a different version than `@latest` impossible (e.g., `:LspInstall gopls@2.0.0`). We'd need to default to `latest` if a specific version has not been requested (`grep requested_server_version` might give some more context). Also, it seems like `go install` really doesn't like providing multiple packages, so I think we should take the opportunity of changing the `go.packages -> go.package` and make it receive a single package name instead of a table. I'll look into this as soon as I'm back at the 💻, otherwise feel free to look into this if you're interested!

smgt:
@williamboman After taking a closer look (i'm not familiar with either lua or *vim plugins) I see the flaws in my PR. I now also understand better how nvim-lsp-installer works. I started to play around with a new PR what you have suggested in your comment. But I don't know how long it will take me to figure everything out.

williamboman:
Superseded by #438, thanks both!

PR #441, opened 2022-01-23
add golangci_lint_ls

williamboman:


samodostal:
This language server isn't working for me. It is attached to the buffer, but I doesn't show any diagnostics. I tried playing with the default values / reinstalling the golangci-lint-langserver command, but I can't figure it out

williamboman:
@samodostal I'm afraid I'm not particularly familiar with this server (or Go in general). Does the project you're working on have a golangci-lint config file? I've tested this using the https://github.com/golangci/golangci-lint repo itself, and it works fine for me:

Screenshot 2022-02-05 at 15 33 29

samodostal:
I just tried it in the repo and still no diagnostics. I'll try to play a bit more with the default_options. This is a screenshot of the language server attached. Maybe I am missing something?
image

williamboman:
Are you setting up the server directly via lspconfig like `require("lspconfig").golangci_lint_ls.setup {}`? If that's the case, it won't really work with nvim-lsp-installer as it assumes things exist on PATH already (see the README for an [example setup snippet](https://github.com/williamboman/nvim-lsp-installer#setup))

samodostal:
I have it set up with the `lsp_installer.on_server_ready`. I install all language servers through `:LspInstallInfo` and all are working correctly for me, except this one

williamboman:
Hmm.. and when you say you're not getting any diagnostics, are you sure it should be giving any diagnostics? Did you add an unused variable, or something similar?

samodostal:
Yes, I tried the `var foo = "bar"` and also if statement that should be changed to switch

williamboman:
Did you write the file as well? I noticed it didn't really seem too keen on applying diagnostics on unsaved changes

samodostal:
Yes I did. I will try to set it up without nvim-lsp-installer, and see what could cause the problem

samodostal:
Okay I fixed it. I looked at logs and saw that there were some warnings about missing packages `gh` and `wget` (I reinstalled brew recently for a newer Mac architecture, so I didn't realise there were some packages missing). And after reinstalling the `golangci_lint_ls` language server, I can see diagnostics correctly.
The log warnings are weird, because they also contained `installation completed, success=true` for the language server. Maybe there are some missing dependencies for the language server?

Anyways, thanks you for your help :)

williamboman:
Cool :+1: `gh` and `wget` are actually both optional (or well, as long as you have `curl`), and I don't believe golangci_lint_ls makes any use of those as it installs everything via `go install`. But good to hear it works now!

PR #455, opened 2022-02-01
[FEATURE] Winrar support for windows.

Borwe:
Since I assume winrar is most common user installed extractor on windows, it would be good as an option for the windows exctrators, can even save user time from having to install the others, which don't normally come with windows out of the box.

[Discovery]
Tested it with rust lsp install, since I just installed windows, and didn't have the other options installed but installed winrar as is normal for me, thought this would be a great addition.

Winrar also supports extracting many more types of archives, could be a big bonus.


williamboman (commented):
Hello! Cool! I remember trying to add this before, but if I remember correctly the command would just open the GUI app - it'd not actually block and suspend until it's actually done unpacking the archive. Maybe I was using a different version?

Borwe:
Well my current version didn't open any gui with the command: ```winrar e file.gz```

williamboman:
> Well my current version didn't open any gui with the command: `winrar e file.gz`

Awesome!

williamboman (approved):

PR #467, opened 2022-02-08
Add r_language_server

meriops:
Add 'r_language_server' in a self-contained way. All credit goes to @aspeddro for getting me started with his fork, and to @josefhajnala for [languageserversetup](https://github.com/jozefhajnala/languageserversetup). 
There might be a more elegant way to do this (because as the saying goes, "I have no idea what I'm doing"), but it seems to work for me here. The initial install can be quite long because the dependencies of `languageserver` are brought along if they are not already present on the system.


williamboman:
Cool! I'm running into errors locally, seemingly because it tries to write to the final destination directory before it's created:

```
        R version 4.1.2 (2021-11-01) -- "Bird Hippie"
        Copyright (C) 2021 The R Foundation for Statistical Computing
        Platform: aarch64-apple-darwin20.6.0 (64-bit)
        
        R is free software and comes with ABSOLUTELY NO WARRANTY.
        You are welcome to redistribute it under certain conditions.
        Type 'license()' or 'licence()' for distribution details.
        
          Natural language support but running in an English locale
        
        R is a collaborative project with many contributors.
        Type 'contributors()' for more information and
        'citation()' on how to cite R or R packages in publications.
        
        Type 'demo()' for some demos, 'help()' for on-line help, or
        'help.start()' for an HTML browser interface to help.
        Type 'q()' to quit R.
        
        > options(langserver_library = "/Users/williamboman/.local/share/nvim/lsp_servers/r_language_server");rlsLib = getOption("langserver_library");install.packages("languageserversetup", lib = rlsLib);loadNamespace("languageserversetup", lib.loc = rlsLib);languageserversetup::languageserver_install(
        Warning in install.packages("languageserversetup", lib = rlsLib) :
          'lib = "/Users/williamboman/.local/share/nvim/lsp_servers/r_language_server"' is not writable
        Error in install.packages("languageserversetup", lib = rlsLib) : 
          unable to install packages
        Execution halted
```

The default behavior is that all installations run in a temporary directory (accessed via `ctx.install_dir` in the `ServerInstallerFunction`), and upon successful completion are moved to the final installation directory. Would it be possible to do the same here?

williamboman:
And if I manually create the final directory myself, it seems to get stuck on:

```
        > options(langserver_library = "/Users/williamboman/.local/share/nvim/lsp_servers/r_language_server");rlsLib = getOption("langserver_library");install.packages("languageserversetup", lib = rlsLib);loadNamespace("languageserversetup", lib.loc = rlsLib);languageserversetup::languageserver_install(
        Error in contrib.url(repos, type) : 
          trying to use CRAN without setting a mirror
        Calls: install.packages -> startsWith -> contrib.url
        Execution halted
```

meriops:
Indeed, I also encountered problems after the push... Sorry about that. 
By default, R packages are already installed in a temporary folder, which is usually `/tmp/`. I think I can use another location as the temporary install directory from the R side, but I need a bit of help with the lua side of things... 
Here, I considered `root_dir` to be the target install directory (ie where all the LSPs end up being installed). However, should `root_dir` point to the _temporary_ install directory instead? And/or what would be the best way to pass `install_dir` (ie, the temporary directory) to the installer function, ie to `process.spawn('R', ... )` ?

williamboman:
> Here, I considered `root_dir` to be the target install directory (ie where all the LSPs end up being installed). However, should `root_dir` point to the _temporary_ install directory instead? And/or what would be the best way to pass `install_dir` (ie, the temporary directory) to the installer function, ie to `process.spawn('R', ... )` ?

Yeah this code has gotten a bit messy over time. What `root_dir` refers to is the final installation dir of the server - the only assumption that can be made is that after successful completion of the installation steps, this directory will exist containing the installation artifacts. This is generally useful if you need to persist this directory in some build artifact.

To access the temporary directory in which the installation actually takes place, you can do so by accessing the `install_dir` in the installer context, like so:

```lua
function (_, callback, ctx)
    process.spawn('R', {
       -- Maybe it'd make sense to run the R program with the working directory set to the temporary install dir?
       -- I assume that the R program could then just use the `.` path as the installation target - effectively targeting the working directory
       cwd = ctx.install_dir, 
       ...
    }, callback)
end
```

williamboman:
@meriops I opened a PR to your branch in your fork [here](https://github.com/meriops/nvim-lsp-installer/pull/1), with some stylistic suggestions.

williamboman:
@meriops I see that the `languageserversetup::languageserver_install()` function takes a `Ncpus` argument which defaults to `max(1L, getOption("Ncpus"))`. Would it make sense to configure this as well, or does the `getOption("Ncpus")` take care of that automatically?

williamboman:
It takes about 40 minutes to compile this on an old, low end, MacBook. Mac M1 (ARM) doesn't seem to work (I believe due to https://github.com/r-lib/fs/issues/348).

meriops:
> @meriops I see that the `languageserversetup::languageserver_install()` function takes a `Ncpus` argument which defaults to `max(1L, getOption("Ncpus"))`. Would it make sense to configure this as well, or does the `getOption("Ncpus")` take care of that automatically?

I believe many R users installing packages from sources would have Ncpus configured in their R startup files, so I would leave that out.

williamboman (approved):
LGTM!

meriops:
> It takes about 40 minutes to compile this on an old, low end, MacBook. Mac M1 (ARM) doesn't seem to work (I believe due to [r-lib/fs#348](https://github.com/r-lib/fs/issues/348)).

Sorry about this, I don't have access to a mac to check. If only the base set of R packages was previously installed on the machine, then the first install can be quite long because the `languageserver` package is pulling a few "heavy" dependencies (`fs` being one of those). The `fs` issue you mention seemed to be fixed (https://github.com/r-lib/fs/pull/365) but not yet merged.

yech1990:
Hi @meriops, do you have a similar issue in installing r_language_server

```bash
The install-github.me service is deprecated, please stop using it.                                                                                                                                          Downloading GitHub repo REditorSupport/languageserver@master                                                                                                                                                                                                                       
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory                                                                                                                                                                
/usr/bin/gtar: .: Cannot getcwd: No such file or directory                                                                                                                                                                                                                         
/usr/bin/gtar: Error is not recoverable: exiting now                                                                                                                                                                                                                               
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory                                                                                                                                                                
Error: Failed to install 'languageserver' from GitHub:                                                                                                                                                                                                                             
  Does not appear to be an R package (no DESCRIPTION)                                                                                                                                                                                                                              
In addition: Warning message:                                                                                                                                                                                                                                                      
In utils::untar(tarfile, ...) :                                                                                                                                                                                                                                                    
  ‘/usr/bin/gtar -xf '/home/user/RtmpcxzwqS/file119de930707d94.tar.gz' -C '/home/user/RtmpcxzwqS/remotes119de94f8d94a'’ returned error code 2                                                                                                                                        
Execution halted                                                                                                                                                                                                                                                                   
shell-init: error retrieving current directory: getcwd: cannot access parent directories: No such file or directory

``

meriops:
@yech1990 Indeed, I had that at some point. The install of `r_language_server` here is set so that if some required packages are already present in `R/library`, they are not installed again. But it turns out that some of these packages (for me it was `desc` and a couple others ; a grep on the R/library folder will tell you) still had the `install-github.me` URL hard-coded. I simply removed and reinstalled them.
@williamboman To avoid this altogether, we might set `fullReinstall = TRUE` and `strictLibrary = TRUE` in the install script here. I initially set those to FALSE in the hope of speeding up the install of `r_language_server`.

yech1990:
Thank you very much @meriops. Is there a option to set `fullReinstall = TRUE` without forking the plugin?

yech1990:
thank you for the information. I still have the some problem in one of the machine, even if "fullReinstall" is True.

meriops:
William: indeed, it's already set. I got confused for a minute. 
y9c: the problem seems to be related to the tar utility which cannot find the parent directory (ie the Rtmp folder was removed along the way for some reason) and/or the remotes R package (see issue 350 there for a somewhat similar situation), rather than to lsp-installer or r_language_server.

scjlee:
> @yech1990 Indeed, I had that at some point. The install of `r_language_server` here is set so that if some required packages are already present in `R/library`, they are not installed again. But it turns out that some of these packages (for me it was `desc` and a couple others ; a grep on the R/library folder will tell you) still had the `install-github.me` URL hard-coded. I simply removed and reinstalled them.

It is probably due to the older version of `languageserversetup` hosted on CRAN. I modified install script to use latest (0.1.2.900) branch on GitHub and the installation just work fine. @williamboman do you want me open a PR?

williamboman:
> It is probably due to the older version of `languageserversetup` hosted on CRAN. I modified install script to use latest (0.1.2.900) branch on GitHub and the installation just work fine. @williamboman do you want me open a PR?

Sure! I can't say I'm following exactly what the problem is, but PRs are always welcomed!

PR #472, opened 2022-02-10
add omnisharp net6.0 support instead of mono

iabdelkareem:
This PR will add support for omnisharp language srever that runs on .NET6.0 instead of mono, the support for .NET6.0 omnisharp came out in the release `1.38.0` and will replace the mono version as mentioned in this PR `https://github.com/OmniSharp/omnisharp-roslyn/pull/2291#issuecomment-985834029`

### DIFF
--- a/lua/nvim-lsp-installer/servers/omnisharp/init.lua
+++ b/lua/nvim-lsp-installer/servers/omnisharp/init.lua
@@ -14,6 +14,12 @@ return function(name, root_dir)
         homepage = "https://github.com/OmniSharp/omnisharp-roslyn",
         languages = { "c#" },
         installer = {
+            std.ensure_executables {
+                {
+                    "dotnet",
+                    "dotnet was not found in path. Refer to https://dotnet.microsoft.com/download for installation instructions.",
+                },
+            },
             context.use_github_release_file(
                 "OmniSharp/omnisharp-roslyn",
                 coalesce(
--- a/lua/nvim-lsp-installer/servers/omnisharp/init.lua
+++ b/lua/nvim-lsp-installer/servers/omnisharp/init.lua
@@ -4,7 +4,6 @@ local path = require "nvim-lsp-installer.path"
 local Data = require "nvim-lsp-installer.data"
 local std = require "nvim-lsp-installer.installers.std"
 local context = require "nvim-lsp-installer.installers.context"
-local dotnet = require "nvim-lsp-installer.installers.dotnet"
 
 local coalesce, when = Data.coalesce, Data.when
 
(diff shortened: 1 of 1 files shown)


lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -17,38 +18,40 @@ return function(name, root_dir)
...
         },
         default_options = {
             cmd = {
-                platform.is_win and path.concat { root_dir, "omnisharp", "OmniSharp.exe" } or path.concat {
-                    root_dir,
-                    "omnisharp",
-                    "run",
-                },
+                "dotnet",
+                path.concat { root_dir, "OmniSharp.dll" },
```
williamboman, line 54: I believe the following fixes the issues for the Mac archive (we'll have to make sure the structure of each .zip archive is the same though):

```suggestion
                path.concat { root_dir, "omnisharp", "OmniSharp.dll" },
```

williamboman: Ah so all the zip files are luckily of the same structure. The reason `omnisharp` is needed here is because the [unzip is currently instructed to unzip to the `omnisharp` directory](https://github.com/williamboman/nvim-lsp-installer/blob/455de5e7a608b830b1756403b73322d8844b50f6/lua/nvim-lsp-installer/servers/omnisharp/init.lua#L45)

lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -4,6 +4,7 @@ local path = require "nvim-lsp-installer.path"
 local Data = require "nvim-lsp-installer.data"
 local std = require "nvim-lsp-installer.installers.std"
 local context = require "nvim-lsp-installer.installers.context"
+local dotnet = require "nvim-lsp-installer.installers.dotnet"
```
williamboman, line 7:
```suggestion

```

lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -17,38 +18,40 @@ return function(name, root_dir)
             context.use_github_release_file(
```
williamboman, line 18: Seeing as the server no longer is a standalone executable and instead runs via the dotnet SDK, I think it'd be nice to make sure it's available on the system (although it's not necessary to just install it):

```suggestion
            std.ensure_executables {
                {
                    "dotnet",
                    "dotnet was not found in path. Refer to https://dotnet.microsoft.com/download for installation instructions.",
                },
            },
            context.use_github_release_file(
```

lua/nvim-lsp-installer/servers/omnisharp/init.lua
iabdelkareem: Thanks for that! I didn't notice while testing that I had the path hardcoded in my config. That's why it didn't pick up the default from here.

lua/nvim-lsp-installer/servers/omnisharp/init.lua
iabdelkareem: Thanks ^^

Issue #334, opened 2021-12-14
Cannot access configuration for emmet-language-server | attempt to index field 'manager' (a nil value)


Elysium1436:
Hi, i'm trying a custom language server, "@kozer/emmet-language-server", and it was working fine until I formatted my pc and put back the exact same config. Once I have it installed with nvim-lsp-installer (which shows up in LspInstall), this message appears:
```
[lspconfig] Cannot access configuration for emmet-language-server. Ensure this server is listed in `serv
er_configurations.md` or added as a custom server.                                                      
[lspconfig] Cannot access configuration for emmet-language-server. Ensure this server is listed in `serv
er_configurations.md` or added as a custom server.                                                      
[nvim-lsp-installer] ...ged/nvim-lsp-installer/lua/nvim-lsp-installer/server.lua:83: attempt to index fi
eld 'manager' (a nil value) 
```

Here is the rc that registers the server
```lua
local lspconfig = require "lspconfig"
local configs = require "lspconfig/configs"
local servers = require "nvim-lsp-installer.servers"
local server = require "nvim-lsp-installer.server"
local path = require "nvim-lsp-installer.path"
local npm = require "nvim-lsp-installer.installers.npm"
local server_name = "emmet-language-server"





configs[server_name] = {
    default_config = {
      filetypes = {
          'html', 'typescriptreact', 'javascriptreact', 'javascript',
          'typescript', 'javascript.jsx', 'typescript.tsx', 'css'
      },
      root_dir = lspconfig.util.root_pattern ".git",
    };
}


local root_dir = server.get_server_root_path(server_name)


-- You may also use one of the prebuilt installers (e.g., std, npm, pip3, go, gem, shell).
local my_installer = npm.packages{'@kozer/emmet-language-server'}

-- 2. (mandatory) Create an nvim-lsp-installer Server instance
local my_server = server.Server:new {
    name = server_name,
    root_dir = root_dir,
    installer = my_installer,
    default_options = {
        cmd = {path.concat{ root_dir,"node_modules",".bin", "emmet-language-server" }, "--stdio" },
    },
}

-- 3. (optional, recommended) Register your server with nvim-lsp-installer.
--    This makes it available via other APIs (e.g., :LspInstall, lsp_installer.get_available_servers()).
servers.register(my_server)
```

Here is the LspInstaller rc
```lua

local lsp_installer = require("nvim-lsp-installer")

lsp_installer.on_server_ready(function(server)
    local opt = {}

[...]

PR #488, opened 2022-02-17
Fix configs require in custom server example

rentziass:
As mentioned in #334, `lspconfig` recently changed the way configs are required. This updates the full example in custom server documentation.


williamboman (approved):
Ah, thanks!

PR #491, opened 2022-02-18
Adjust docs and installer link for jsonnet-language-server

zalegrala:


zalegrala:
I think they just moved the repo.

PR #492, opened 2022-02-19
Add slint language server for .slint files

hunger:
https://slint-ui.com/ has the details to the language and LS implementation.

The binary is called slint_lsp (which is also the name in nvim-lspconfig), so I would like to keep that here as well.

### DIFF
--- a/lua/nvim-lsp-installer/servers/slint_lsp/init.lua
+++ b/lua/nvim-lsp-installer/servers/slint_lsp/init.lua
@@ -30,7 +30,7 @@ return function(name, root_dir)
         },
         default_options = {
             cmd_env = {
-                 PATH = process.extend_path { path.concat { root_dir, "slint-lsp" } },
+                PATH = process.extend_path { path.concat { root_dir, "slint-lsp" } },
             },
         },
     }
--- a/lua/nvim-lsp-installer/servers/slint_lsp/init.lua
+++ b/lua/nvim-lsp-installer/servers/slint_lsp/init.lua
@@ -1,0 +1,37 @@
+local server = require "nvim-lsp-installer.server"
+local path = require "nvim-lsp-installer.path"
+local Data = require "nvim-lsp-installer.data"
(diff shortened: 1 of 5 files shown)


williamboman (commented):
Cool!! I'm however getting the following error when trying to run the binary on an Ubuntu 20.04 machine (x64):

```
./slint-lsp: error while loading shared libraries: libQt5Core.so.5: cannot open shared object file: No such file or directory
```

lua/nvim-lsp-installer/servers/slint_lsp/init.lua
```diff
@@ -0,0 +1,36 @@
...
+                    return std.unzip_remote(ctx.github_release_file)
+                else
+                    return std.untargz_remote(ctx.github_release_file)
+                end
+            end),
+        },
+        default_options = {
+            cmd_env = {
+                PATH = process.extend_path { root_dir, "slint-lsp" },
+            },
```
williamboman, line 33: By extending the PATH environment used when starting the server, we can avoid having to override the `cmd` set by lspconfig.

williamboman (approved):
Great work, LGTM!

hunger:
> Cool!! I'm however getting the following error when trying to run the binary on an Ubuntu 20.04 machine (x64):
> 
> ```
> ./slint-lsp: error while loading shared libraries: libQt5Core.so.5: cannot open shared object file: No such file or directory
> ```

Those libraries are taken from the system and not part of the linux tarball. So I guess you just do not have those installed?

lua/nvim-lsp-installer/servers/slint_lsp/init.lua
hunger: Hey, thanks for fixing this!

hunger:
One sec, give a a chance to test:-)

hunger:
Ok, works for me again, now using PATH! Thanks for the patch putting me on the right track!

Issue #510, opened 2022-02-27
LspInstallInfo should show server version


cjnucette:
### Is your feature request related to a problem? Please describe.

It is cumbersome to find out which version of the server is installed.

### Describe the solution you'd like

Show version alongside the name of the server in the LspInstallInfo window.

williamboman:
Hello! Yes this is definitely something I want to do. It should actually be pretty straight-forward as everything is mostly in place, will look into soon. Just out of curiosity - is there some specific reason why you want/need to do this?

williamboman:
So this wasn't nearly as straight forward to do as I had first thought, but here's a sneak peak:

https://user-images.githubusercontent.com/6705160/156441021-62d049e1-6199-4db1-b286-fafa24726dd7.mov

cjnucette:
I was trying to find out which version of `sumneko_lua` was installed, because of the new formatting feature added to it, and I wanted to test it.
Thank you.

williamboman:
See branch `ui/display-current-version` (#520). Just gotta do some cleaning up now

PR #512, opened 2022-03-01
Add Dhall LSP

sondr3:


lua/nvim-lsp-installer/servers/dhall_lsp_server/init.lua
```diff
@@ -0,0 +1,43 @@
...
+local std = require "nvim-lsp-installer.installers.std"
+local context = require "nvim-lsp-installer.installers.context"
+local Data = require "nvim-lsp-installer.data"
+
+local coalesce, when = Data.coalesce, Data.when
+
+local target = coalesce(
+    when(platform.is_mac, "dhall-lsp-server-1.0.18-x86_64-macos.tar.bz2"),
+    when(platform.is_linux, "dhall-lsp-server-1.0.18-x86_64-linux.tar.bz2"),
+    when(platform.is_win, "dhall-lsp-server-1.0.18-x86_64-windows.zip")
```
sondr3, line 14: Really not sure how to handle this, Dhall releases the tools and binaries alongside language updates and they are not versioned together. So for the latest release of the Dhall language (`1.41.1`), the LSP is at version `1.0.18` but was at `1.0.17` for `1.40.0`. Any suggestions for ways to handle this?

williamboman: Ah yes I remember this one actually. I was working on adding Dhall some time ago but ran into this and gave up. It's a pretty odd way of distributing releases. Had they produced separate releases (a separate tag) for the LSP server versions this would've been very easy to support as-is (`:Ggrep tag_name_pattern` in nvim-lsp-installer source), but now they just bundle everything together under a single release which complicates things - we'll need to manually scan the release assets provided in the GitHub API response.

The long answer (sorry): The current way of composing together different "installer functions" (e.g. `context.use_github_release_file()`, `context.capture()`) has turned out to be 90% of the time just a poor man's implementation of async/await (which Lua lacks natively, but is somewhat trivial to implement via Lua coroutines). The remaining 10% are cases where treating an installer function as a distinct, composable, unit actually makes sense. For these reasons I recently added an [async module](https://github.com/williamboman/nvim-lsp-installer/pull/499), which I intend to make a broad switch to over time. With this in place, it'll become easier to compose asynchronous operations without having to conform to the somewhat static installer functions, something like (pseudo-ish code):

```lua
function (context)
    local latest_release = github.fetch_latest_release("dhall-lang/dhall-haskell") -- suspends until it resolves
    local release_asset = list_find_first(function (asset)
         return asset.match(string.format("dhall-lsp-server-.+-x_86_64-%s", current_platform))
    end, latest_release.assets)
    assert(release_asset, "Some nice error message here.")
    download_file(release_asset) -- suspends until it resolves
    -- ...
end
```

The above could of course be further abstracted ad infinitum, but at the end of the day it's all just normal Lua functions running in a coroutine context that allows for suspending until some async operation resolves.

I'd be happy to start exploring the above ideas in order to enable Dhall support! I'm currently preoccupied with #510, but I'll look into this PR immediately after.

williamboman (commented):
Awesome!

lua/nvim-lsp-installer/servers/dhall_lsp_server/init.lua
williamboman: What we could do right away however is to pin this server to a specific version (i.e. release [`1.41.1`](https://github.com/dhall-lang/dhall-haskell/releases)  and target `dhall-lsp-server-1.0.18*`)

sondr3: I thing it's a fine temporary fix, I assume all that is required is changing `target` to the version we want to pin?

williamboman: Yep exactly. We'll also need to pin the GitHub release version, what you could do is put the following in the very beginning of the installation (otherwise see `rome/init.lua` for a similar case of pinning a version):

```lua
            context.set(function(ctx)
                ctx.requested_server_version = Data.coalesce(
                    ctx.requested_server_version,
                    "1.41.1"
                )
            end),
```

williamboman:
Thanks @sondr3!

PR #518, opened 2022-03-04
remove npm install --production from ansiblels

joshzcold:
Latest ansible-language-server fails on install with this line present.
the combination of `npm install` and `npm run compile` should be enough since `npm install` gets both the production and dev dependencies.  

`npm install --production` fails with missing typescript type definitions missing. 
```
 node_modules/@nodelib/fs.scandir/out/providers/async.d.ts(1,23): error TS2688: Cannot find type definition file for 'node'.
 node_modules/@nodelib/fs.scandir/out/providers/async.d.ts(4,45): error TS2503: Cannot find namespace 'NodeJS'.
 node_modules/@nodelib/fs.scandir/out/types/index.d.ts(1,23): error TS2688: Cannot find type definition file for 'node'.
```

### DIFF
--- a/lua/nvim-lsp-installer/servers/ansiblels/init.lua
+++ b/lua/nvim-lsp-installer/servers/ansiblels/init.lua
@@ -15,7 +15,6 @@ return function(name, root_dir)
             npm.install { "npm@latest" }, -- ansiblels has quite a strict npm version requirement
             npm.exec("npm", { "install" }),
             npm.run "compile",
-            npm.exec("npm", { "install", "--production" }),
             context.receipt(function(receipt)
                 receipt:with_primary_source(receipt.git_remote "https://github.com/ansible/ansible-language-server")
             end),


treatybreaker:
I can back this change, just did this myself during lunch then noticed this pull request before I made mine 😄 . Since we're building from source we do need the typescript dependencies. Seems to be related to [this commit](https://github.com/ansible/ansible-language-server/commit/1366891c9f662d7dc9f59fcea46175bd724d5892) from `ansiblels`. Using a revision before that commit works fine -- arguably an `ansiblels` issue instead of this project.

williamboman (approved):
Ah, thanks!

joshzcold:
BTW @williamboman there is a current issue with ansible-language-server that I have a pull request for that effects neovim users: https://github.com/ansible/ansible-language-server/pull/246

Is there a good way to configure to use my fork with lsp-installer before this pull request gets merged? 
Im currently editing `~.local/share/nvim/site/pack/packer/start/nvim-lsp-installer/lua/nvim-lsp-installer/servers/ansiblels/init.lua`
to accomplish this.

PR #534, opened 2022-03-11
chore(volar): update volar package name

sethidden:
see https://github.com/johnsoncodehk/volar/issues/876
see https://github.com/johnsoncodehk/volar/blob/master/packages/vue-language-server/package.json#L2


sethidden:
moved to draft since the @volar/vue-language-server package is not published yet

williamboman (approved):
Thank you!

PR #599, opened 2022-04-15
Update README.md

hermitmaster:
Missing cargo from the list of requirements to install all servers.


williamboman (approved):
Ah, cool. Will add this to healthcheck and `doc/nvim-lsp-installer.txt` as well!

PR #608, opened 2022-04-18
Fix the warning of the lua_language_server

chen244:
Try to fix the warning below
image


williamboman:
Hey! Thanks for the PR! sumneko_lua used to be able to find these metatable methods automatically, but it seems like it stopped working somewhere after the 3.0.0 release. I currently run `sumneko_lua@v2.6.7` because I'm experiencing some other type system issues as well, on that version this is not a problem!

I _believe_ this is an actual issue upstream with sumneko_lua, so I'll hold on merging this for now in the hopes that it'll be resolved upstream instead!

PR #618, opened 2022-04-21
add deprecation notice to old installer modules

williamboman:


williamboman:
@kylo252 👋 I've done some significant overhauls recently, where as a result these modules are now deprecated and should not be used. I've written a short doc describing the changes here: https://github.com/williamboman/nvim-lsp-installer/wiki/Async-infrastructure-changes-notice.

I don't believe LunarVim uses any custom servers in core, but there might be some users who registers custom server installers who might be presented with this deprecation notice once LunarVim bumps the plugin version.

williamboman:
> thanks for the heads-up! mind creating an issue to track this, which could also be pinned, for better visibility?

I created a discussion instead: https://github.com/williamboman/nvim-lsp-installer/discussions/619. Feel free to send people with questions there, although I 1) I think a very small portion of users use custom servers, and 2) those who do are resourceful enough to figure it out themselves :D

Issue #264, opened 2021-11-17
esbonio: Need install all plugins of sphinx for esbonio


cathaysia:
Hi, had remember that bug I mention? https://github.com/williamboman/nvim-lsp-installer/pull/253

I had talk with https://github.com/swyddfa/esbonio/issues/45 , then I solve this. In the short, for a spinx project, users my import many plugins, we **must** install all plugins for esbonio, else go dead. :(

So I want know how do we solve this?

source ~/.local/share/nvim/lsp_servers/esbonio/venv/bin/activate then install requirements.txt ?

Or can we change plugins that need be installed in user's configuration?
And or we remove esbonio installer?

Looking forward to your replay. And sorry for my pr cause this bug. :(

williamboman:
Hello! Do I understand you correctly that it's common to declare project-local Sphinx dependencies in `requirements.txt`, which will need to be installed in the same venv as Esbonio?

I guess what could be done is to add a new command like `:EsbonioInstallPlugins`, which will use the same venv as the esbonio installation. This command would install packages defined in `project_root_dir .. "/requirements.txt"`. If possible this should probably also install to a different `PIP_TARGET` to avoid polluting the esbonio installation itself. Not sure if this makes sense, but I feel like it should be doable?

cathaysia:
Yes, it's all I want express. :)

For my part, I declare all dependencies in requirements.txt, I can't sure others whether do this. So there are two ways. either write docs to tell users install all dependencies by themself, or add `:EsbonioInstallPlugins` commands.

I'm not familiar with PIP_TARGET, but I think, emm, dependencies shouldn't polluting esbonio itself. Because sphinx's plugins don't conflict with each other.(I haven't encounter this problem.) So maybe we can install all dependencs of all projects in a venv.

And, sorry for my plastic English.

wainscape:
If I am understanding correctly, I had a similar issue. I want to take advantage of Pylsp's plugin support to use the Black formatter. Although Pylsp natively supports several third-party tools, such as autoPEP8, if they're installed, at this time Black is supported via a third party plugin, [python-lsp-black](https://github.com/python-lsp/python-lsp-black). 

To install the black plugin:
`source ~/.local/share/nvim/lsp_servers/pylsp/venv/bin/activate`
`pip install python-lsp-black`

And now it works. However, it could be handy to have some integration with the LspInstall commands and interface, which are so wonderfully convenient and easy to use. 

[...]

PR #623, opened 2022-04-22
feat(esbonio): add command EsbonioInstallPlugins

cathaysia:
It's time to solve this bug: https://github.com/williamboman/nvim-lsp-installer/issues/264

but I have not test this, because I am not using lsp_installer.on_server_ready(function(server) to setup lsps, and I don't know how to setup commands by myself :(

### DIFF
--- a/lua/nvim-lsp-installer/servers/esbonio/init.lua
+++ b/lua/nvim-lsp-installer/servers/esbonio/init.lua
@@ -1,5 +1,7 @@
 local server = require "nvim-lsp-installer.server"
 local pip3 = require "nvim-lsp-installer.core.managers.pip3"
+local process = require "nvim-lsp-installer.process"
+local notify = require "nvim-lsp-installer.notify"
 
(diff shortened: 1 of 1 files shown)


williamboman (commented):
Hey! Hmm I'm not too sure about this to be honest (not the code but the feature itself). I feel like if you rely on packages being available in the very same venv that esbonio is running in, it'd be much better if you simply make sure to install esbonio in that venv instead of carrying over all the dependencies into the nvim-lsp-installer venv.

I also feel like this would carry questionable support for multiple projects (what happens if you run `:EsbonioInstallPlugins` in say 5 different projects with conflicting dependencies)?

Another thing this would carry is that it'd compromise the integrity of the nvim-lsp-installer venv, by fiddling with its installed packages post-installation.

cathaysia:
Yes, you are right. Install esbonio's dependence in venv is a better choice. I have not give it a good consider. Sorry.

Issue #624, opened 2022-04-23
Failed to install `rust-analyzer`


Elio-zhy:
### Problem description

When I tried to install `rust-analyzer`, I got `Unable to pack rust-analyzer.exe` error and it failed.
And it also gave me `The system cannot find the file specified` error.
![image](https://user-images.githubusercontent.com/30508272/164883403-57a0435b-d40a-45d8-8eb2-044becc9259e.png)

### Neovim version (>= 0.6)

NVIM v0.7.0
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3

### Operating system/version

Windows 10 Pro 21H2

### I've recently downloaded the latest plugin version of both nvim-lsp-installer and nvim-lspconfig

- [X] Yes

### Affected language servers

rust-analyzer

### Actual behavior

Installation of `rust-analyzer` failed with `Unable to unpack rust-analyzer.exe.` error.

### Expected behavior

Install without error.

### LspInstallInfo output

```shell
Installed servers (3) 
    ◍ clangd (c, c++)
    ◍ pylsp (python)
    ◍ sumneko_lua (lua)
  
  Pending servers (1) 
    ◍ rust_analyzer (failed)
        Downloading file "https://github.com/rust-lang/rust-analyzer/releases/download/2022-04-18/rust-analyzer-x86_64-pc-windows-msvc.gz"...
        
        7-Zip 21.07 (x64) : Copyright (c) 1999-2021 Igor Pavlov : 2021-12-26
        
        Scanning the drive for archives:
        
        ERROR: The system cannot find the file specified.
        rust-analyzer.exe
        
        
        
        System ERROR:
        The system cannot find the file specified.
        Could not find executable "peazip" in path.
        Could not find executable "wzunzip" in path.
        Could not find executable "winrar" in path.
        ...taller/lua/nvim-lsp-installer/core/managers/std/init.lua:102: Unable to unpack rust-analyzer.exe.
```

### Installation log

```shell
[DEBUG 2022/4/23 15:02:41] ...-installer/lua/nvim-lsp-installer/ui/status-win/init.lua:729: Queuing server="rust_analyzer", version="" for installation
[INFO  2022/4/23 15:02:41] ...-installer/lua/nvim-lsp-installer/ui/status-win/init.lua:682: Starting install server_name="rust_analyzer", requested_version=""
[...]

PR #625, opened 2022-04-23
fix(rust-analyzer): unzip filename error

Elio-zhy:
This is a solution for #624 .
I'm not sure if it's the real problem but it does solve my problem.


williamboman (approved):
Ah thanks!

PR #629, opened 2022-04-25
Add VLS (for vlang)

ozgb:
https://github.com/vlang/vls

https://vlang.io/

- Requires that V is already installed and on the PATH

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,6 +291,7 @@ require("nvim-lsp-installer").setup({
 | Terraform                           | `terraformls`              |
 | Terraform [(docs)][tflint]          | `tflint`                   |
 | TypeScript                          | `tsserver`                 |
+| V                                   | `vls`                      |
 | Vala                                | `vala_ls`                  |
 | VimL                                | `vimls`                    |
 | Vue                                 | `volar`                    |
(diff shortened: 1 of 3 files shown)


williamboman (commented):
Cool! Added a few comments.

Also, the `return function (name, root_dir) ... end` thingy is really confusing, I know. The reason it's needed is because all core installers are instantiated [here](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/init.lua#L216-L220), via the factory function they provide.

oscarbailey-tc:
> Cool! Added a few comments.
> 
> Also, the `return function (name, root_dir) ... end` thingy is really confusing, I know. The reason it's needed is because all core installers are instantiated [here](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/init.lua#L216-L220), via the factory function they provide.

Thanks for the review! I'll update when I get the chance

williamboman:
> Thanks for the review! I'll update when I get the chance

No worries! Let me know if you want me to push these changes directly to this branch if you're unable to look into them!

PR #630, opened 2022-04-26
fix(terraformls): use `root-dir` instead of `root-dir/terraform-ls` as cmd path

mandlm:
The terraform-ls installation contains no subdirectories, the binary gets extracted to the `root-dir` directory.


williamboman (approved):

williamboman:
Ah thanks, this seems to have been inadvertently changed in https://github.com/williamboman/nvim-lsp-installer/pull/616

PR #631, opened 2022-04-26
feat: integrate with lspconfig's on_setup hook

williamboman:
First step towards not requiring any special interactions in user's init scripts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,7 @@ NVIM_HEADLESS:=nvim --headless --noplugin -u tests/minimal_init.vim
 
 dependencies:
 	git clone --depth 1 https://github.com/nvim-lua/plenary.nvim dependencies/pack/vendor/start/plenary.nvim
+	git clone --depth 1 https://github.com/neovim/nvim-lspconfig dependencies/pack/vendor/start/nvim-lspconfig
 
 .PHONY: clean_dependencies
 clean_dependencies:
--- a/README.md
+++ b/README.md
@@ -5,32 +5,31 @@
 -   [About](#about)
 -   [Screenshots](#screenshots)
 -   [Installation](#installation)
-    -   [Packer](#packer)
-    -   [vim-plug](#vim-plug)
 -   [Usage](#usage)
-    -   [Commands](#commands)
     -   [Setup](#setup)
+    -   [Commands](#commands)
     -   [Configuration](#configuration)
 -   [Available LSPs](#available-lsps)
 -   [Custom servers](#custom-servers)
 -   [Logo](#logo)
--   [Roadmap](#roadmap)
 -   [Default configuration](#default-configuration)
 
 ## About
 
-Neovim plugin that allows you to seamlessly install LSP servers locally (inside `:echo stdpath("data")`).
+Neovim plugin that allows you to manage LSP servers (servers are installed inside `:echo stdpath("data")` by default).
+It registers a hook with [`lspconfig`](https://github.com/neovim/nvim-lspconfig) that enhances the `PATH` environment,
+allowing neovim's LSP client to locate the installed server executable<sup>1</sup>.
 
 On top of just providing commands for installing & uninstalling LSP servers, it:
 
 -   provides a graphical UI
--   is optimized for blazing fast startup times
--   provides the ability to check for new server versions
+-   provides the ability to check for, and upgrade to, new server versions through a single interface
 -   supports installing custom versions of LSP servers (for example `:LspInstall rust_analyzer@nightly`)
 -   relaxes the minimum requirements by attempting multiple different utilities (for example, only one of `wget`, `curl`, or `Invoke-WebRequest` is required for HTTP requests)
--   allows you to install and setup servers without having to restart neovim
 -   hosts [a suite of system tests](https://github.com/williamboman/nvim-lspconfig-test) for all supported servers
 -   has full support for Windows <img src="https://user-images.githubusercontent.com/6705160/131256603-cacf7f66-dfa9-4515-8ae4-0e42d08cfc6a.png" height="20">
+
+<sup>1 - some servers don't provide an executable, in which case the full command to spawn the server is provided instead</sup>
 
 ## Screenshots
 
@@ -83,6 +82,17 @@ Plug 'williamboman/nvim-lsp-installer'
 
 ## Usage
 
+### Setup
+
+In order for nvim-lsp-installer to register the necessary hooks at the right moment, **make sure you call the `.setup()`
+function before you set up any servers with `lspconfig`**!
+
+```lua
+require("nvim-lsp-installer").setup {}
+```
+
+Refer to the [Configuration](#configuration) section for information about which settings are available.
+
 ### Commands
 
 -   `:LspInstallInfo` - opens a graphical overview of your language servers
@@ -92,54 +102,16 @@ Plug 'williamboman/nvim-lsp-installer'
 -   `:LspInstallLog` - opens the log file in a new tab window
 -   `:LspPrintInstalled` - prints all installed language servers
 
-### Setup
-
-The recommended way of setting up your installed servers is to do it through nvim-lsp-installer.
-By doing so, nvim-lsp-installer will make sure to inject any necessary properties before calling lspconfig's setup
-function for you. You may find a minimal example below. To see how you can override the default settings for a server,
-refer to the [Wiki][overriding-default-settings].
-
-Make sure you don't also set up your servers directly via lspconfig (e.g. `require("lspconfig").clangd.setup {}`), as
-this will cause servers to be set up twice!
-
-[overriding-default-settings]: https://github.com/williamboman/nvim-lsp-installer/wiki/Advanced-Configuration#overriding-the-default-lsp-server-options
-
-```lua
-local lsp_installer = require("nvim-lsp-installer")
-
--- Register a handler that will be called for each installed server when it's ready (i.e. when installation is finished
--- or if the server is already installed).
-lsp_installer.on_server_ready(function(server)
-    local opts = {}
-
-    -- (optional) Customize the options passed to the server
-    -- if server.name == "tsserver" then
-    --     opts.root_dir = function() ... end
-    -- end
-
-    -- This setup() function will take the provided server configuration and decorate it with the necessary properties
-    -- before passing it onwards to lspconfig.
-    -- Refer to https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md
-    server:setup(opts)
-end)
-```
-
-For more advanced use cases you may also interact with more APIs nvim-lsp-installer has to offer, refer to `:help nvim-lsp-installer` for more docs.
-
 ### Configuration
 
-You can configure certain behavior of nvim-lsp-installer by calling the `.settings()` function.
-
-_Make sure to provide your settings before any other interactions with nvim-lsp-installer!_
+You may optionally configure certain behavior of nvim-lsp-installer when calling the `.setup()` function.
 
 Refer to the [default configuration](#default-configuration) for all available settings.
 
 Example:
 
 ```lua
-local lsp_installer = require("nvim-lsp-installer")
-
-lsp_installer.settings({
+require("nvim-lsp-installer").setup({
     ui = {
         icons = {
             server_installed = "✓",
@@ -289,10 +261,6 @@ You can create your own installers by using the same APIs nvim-lsp-installer its
 
 Illustrations in the logo are derived from [@Kaligule](https://schauderbasis.de/)'s "Robots" collection.
 
-## Roadmap
-
--   Command (and corresponding Lua API) to update outdated servers (e.g., `:LspUpdateAll`)
-
 ## Default configuration
 
 ```lua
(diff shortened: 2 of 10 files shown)


williamboman:
@mjlbach :wave: I'm struggling a bit with the fact that calls to lspconfig's `.setup()` functions usually takes place immediately in user's init scripts. This makes it difficult for nvim-lsp-installer to initialize what it needs at the right moment, without having users manage this either by

1) making sure to call `require('nvim-lsp-installer').setup()` before any lspconfig interactions, or
2) have users leverage the different lifecycle mechanisms that plugin managers like packer provide (`use { "neovim/nvim-lspconfig", after = "lsp_installer" }`)

I was wondering if it'd make sense to somehow defer the setup of each server until the `after/` scripts are executed, leaving room for plugins to run any necessary initialization scripts inside `plugin/`. Or do you have any other ideas?

I'm thinking something like this:

```diff
diff --git a/after/plugin/lspconfig.vim b/after/plugin/lspconfig.vim
new file mode 100644
index 0000000..cf833b7
--- /dev/null
+++ b/after/plugin/lspconfig.vim
@@ -0,0 +1 @@
+lua require('lspconfig.deferred').flush()
diff --git a/lua/lspconfig/configs.lua b/lua/lspconfig/configs.lua
index a47ecba..8e6b4b4 100644
--- a/lua/lspconfig/configs.lua
+++ b/lua/lspconfig/configs.lua
@@ -1,4 +1,5 @@
 local util = require 'lspconfig.util'
+local deferred = require 'lspconfig.deferred'
 local api, validate, lsp = vim.api, vim.validate, vim.lsp
 local tbl_extend = vim.tbl_extend
 
@@ -50,6 +51,11 @@ function configs.__newindex(t, config_name, config_def)
 
     config = tbl_extend('keep', config, default_config)
 
+    if deferred.should_defer then
+      deferred.defer(config)
+      return
+    end
+
     if util.on_setup then
       pcall(util.on_setup, config)
     end
diff --git a/lua/lspconfig/deferred.lua b/lua/lspconfig/deferred.lua
new file mode 100644
index 0000000..0720522
--- /dev/null
+++ b/lua/lspconfig/deferred.lua
@@ -0,0 +1,21 @@
+local M = {}
+
+M.should_defer = true
+
+local deferred_configs = {}
+
+function M.defer(config)
+  assert(M.should_defer, "defer() was called when it wasn't expected.")
+  table.insert(deferred_configs, config)
+end
+
+function M.flush()
+  local configs = require 'lspconfig.configs'
+  M.should_defer = false
+  for _, config in pairs(deferred_configs) do
+    configs[config.name].setup(config)
+  end
+  deferred_configs = {}
+end
+
+return M

```

mjlbach:
Thank you, this is wonderful and addresses my main concern with `nvim-lsp-installer`

`require('nvim-lsp-installer').setup()`

I personally feel like this is reasonable? Unfortunately (and it could be said this is a limitation of the vim package system) we don't have a way to enforce a conditional ordering with neovim.

I'm somewhat ok with the idea of making `setup {}` lazier, but I need to think about it a bit more

williamboman:
I added a mandatory `.setup {}` function, similar to most other plugins.

Also ping @kylo252 @VonHeikemen @junnplus - with this change I'm intending to eject nvim-lsp-installer from being explicitly involved in setting up LSP servers. This will first and foremost target new users of the plugin, but I intend to move over existing users as well (trying to time this with neovim releases to make it feel more natural). The APIs that exist today (`on_server_ready()`, `server:setup()`, `server:on_ready()`, etc.) will remain unchanged, so no worries, but some of them will be deprecated and I hope to fully remove them in the future. I'd be happy to fill in any potential gaps to enable certain use cases, but in that case take the opportunity to redesign them to be less intrusive.

With the emergence of "wrapper/glue" plugins like @VonHeikemen's, @junnplus's et al., I'll also start looking into potentially cutting of other parts of nvim-lsp-installer that aren't integral to the installation and management of LSP servers.

<details>
<summary>An example setup from now on would look something like this:</summary>

```lua
require("nvim-lsp-installer").setup {} -- <- I hope to to be able to even exclude this in the future, but for now it'll be required
local lspconfig = require("lspconfig")

lspconfig.sumneko_lua.setup {
  on_attach = on_attach,
  settings = {
    Lua = {
      format = {
        enable = false
      },
      diagnostics = {
        globals = { "P" },
      },
    }
  }
}

local tsutils = require "nvim-lsp-ts-utils"
lspconfig.tsserver.setup {
  init_options = {
    hostInfo = "neovim",
    preferences = {
      includeInlayParameterNameHints = "none",
      includeInlayParameterNameHintsWhenArgumentMatchesName = false,
      includeInlayFunctionParameterTypeHints = false,
      includeInlayVariableTypeHints = true,
      includeInlayPropertyDeclarationTypeHints = true,
      includeInlayFunctionLikeReturnTypeHints = true,
      includeInlayEnumMemberValueHints = true,
    },
  },
  on_attach = function(client, bufnr)
    common_on_attach(client, bufnr)
    tsutils.setup {}
    tsutils.setup_client(client)
  end,
}

require("rust-tools").setup {}

for _, server in ipairs { "clangd", "eslint", "ltex" } do
  lspconfig[server].setup { on_attach = on_attach }
end
```
</details>

VonHeikemen:
The current implementation of `lsp-zero` should not be affected by this change. I merge the servers default options with the users options and feed it to `lspconfig`. If understand this PR correctly, I don't even need to call the setup function.

Thanks for including me in the conversation.

williamboman:
> It's even trivial currently to get the the cmd_env from nvim-lsp-installer and just pass it directly to lspconfig, e.g. https://github.com/LunarVim/LunarVim/blob/f49f687bc244ff603f3dde6828a85538c50df220/tests/minimal_lsp.lua#L91-L84

Worth noting is that while just extracting cmd_env will work for a majority of servers, there are instances where this won't suffice (for example, [powershell_es](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/powershell_es/init.lua#L21) and [angularls](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/angularls/init.lua#L47-L51)). For these reasons, I think as a principal rule entire tables should always be merged (which happens automatically now if you call `require("nvim-lsp-installer").setup()`).

> I've always preferred this approach, the only blocker at the time was the race condition between the installer still running and lspconfig attempting to launch the server.

This will actually now be the case when using the newly suggested config (going directly via lspconfig). lspconfig will fail to launch configured servers until they're actually installed (this was avoided with the `.on_server_ready()` approach and one of the original motivations behind it). But, all mechanisms around server setup are now being externalized from this plugin, to be solved/improved elsewhere!

---

To those I pinged in my previous comment, after this PR is merged I'd encourage you all to find ways to reduce your dependency on nvim-lsp-installer to one single line: `require("nvim-lsp-installer").setup {}`. If this breaks some functionality or if you struggle with something please ping me somewhere here on GitHub or on IRC (`williamboman` `@libera.chat`) !

williamboman:
Note that if the `require("nvim-lsp-installer").setup {}` function has been called, setting up servers the "old way" via nvim-lsp-installer (`server:setup {}`) will now throw an error. This should be fine as-is, but some users might end up adding that line to their configs while using one of the available auxiliary plugins to set up servers.

Also, I forgot to ping @ray-x with navigator.lua in my comment earlier.

Will merge this now 🤞

axieax:
Hi, just wondering if there is a migration guide for this breaking change for those using `on_server_ready`.

kylo252:
> Worth noting is that while just extracting cmd_env will work for a majority of servers, there are instances where this won't suffice (for example, [powershell_es](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/powershell_es/init.lua?rgh-link-date=2022-04-27T18%3A32%3A41Z#L21) and [angularls](https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/angularls/init.lua?rgh-link-date=2022-04-27T18%3A32%3A41Z#L47-L51)). For these reasons, I think as a principal rule entire tables should always be merged (which happens automatically now if you call `require("nvim-lsp-installer").setup()`).

Thanks for the heads-up! you can see I'm already usign `server:get_default_options()` so it's just a matter of merging the options like you mentioned.

---

Unless I'm missing something, I think you might've forgotten to mention how `lspconfig` would find the server, like read from `server:get_default_options()` or, maybe add the installation root-dir to path.

https://github.com/williamboman/nvim-lsp-installer/blob/110e30b48699531c8e7b82ee08803ae596a88b6b/doc/nvim-lsp-installer.txt#L51-L68

williamboman:
> Hi, just wondering if there is a migration guide for this breaking change for those using `on_server_ready`.

Ah I see some people have interpreted this as a breaking change - there are no breaking changes introduced here! Everything will continue to function as before, there's just a new preferred way of setting things up. I see that one of the commit messages was marked as breaking change and made it through in the squashed commit message, so packer is picking that up as a breaking change when updating the plugin version. I'm drafting an announcement with a guide on how to migrate atm, will publish under Discussions when it's ready!

> Unless I'm missing something, I think you might've forgotten to mention how `lspconfig` would find the server, like read from `server:get_default_options()`, or maybe add the installation root-dir to path.

I'm not sure that needs to be documented, does it? Just call `.setup()`, then you can keep nvim-lsp-installer out of sight & out of mind. Tbh I'm hoping to either completely get rid of the `Server` class, or alternatively just expose `:install()`, `:is_installed()`, and `:uninstall()` as public methods.

kylo252:
> I'm not sure that needs to be documented, does it? Just call `.setup()`, then you can keep nvim-lsp-installer out of sight & out of mind. Tbh I'm hoping to either completely get rid of the `Server` class, or alternatively just expose `:install()`, `:is_installed()`, and `:uninstall()` as public methods.

I did actually originally miss what that the middleware is doing, I see now how it's updating the defaults of lspconfig. 

I'd still say it's worth mentioning that this is what's happening behind the scenes. Especially since `on_setup` is currently undocumented, so we could add the docs for it in lspconfig, and just link to it from here.

image

kylo252:
> Tbh I'm hoping to either completely get rid of the `Server` class, or alternatively just expose `:install()`, `:is_installed()`, and `:uninstall()` as public methods.

on a semi-related note: what's your plans for the integration with `Server:get_settings_schema()`? it looks a bit odd now that the installer isn't directly involved with the setup, but I also know it was quite a bit of effort to get that part up and running smoothly 😢

williamboman:
> I'd still say it's worth mentioning that this is what's happening behind the scenes. Especially since `on_setup` is currently undocumented, so we could add the docs for it in lspconfig, and just link to it from here.

I'm not sure the average user would need to know this though? I feel like it's more geared towards plugin authors? It is actually mentioned briefly in the [introduction in the README](https://github.com/williamboman/nvim-lsp-installer#about)!

> on a semi-related note: what's your plans for the integration with `Server:get_settings_schema()`? it looks a bit odd now that the installer isn't directly involved with the setup, but I also know it was quite a bit of effort to get that part up and running smoothly 😢

I don't consider that method part of the public API, only things documented [in `:h nvim-lsp-installer`](https://github.com/williamboman/nvim-lsp-installer/blob/main/doc/nvim-lsp-installer.txt#L341-L397) are what I consider public, things not included are subject to breaking changes without notice. I will probably at the very least restructure things internally to make sure not to colocate internals with public interfaces. Is this method being used elsewhere currently?

mehalter:
This is an awesome PR and I'm really enjoying the decoupling of the lsp installer and the lsp configuration. I do have a quick clarification question. is the `on_server_ready` function going to be deprecated in the future? Or is that API going to stay in place? It makes it nice to be able to have a function that sets up the servers easily without me needing to add a configuration line simply because I have installed it and it is ready to be used.

williamboman:
> This is an awesome PR and I'm really enjoying the decoupling of the lsp installer and the lsp configuration. I do have a quick clarification question. is the `on_server_ready` function going to be deprecated in the future? Or is that API going to stay in place? It makes it nice to be able to have a function that sets up the servers easily without me needing to add a configuration line simply because I have installed it and it is ready to be used.

It is already deprecated, so continued use is not recommended. As for removal, I don't have any concrete plans. There will probably be tons of configs that still make use of it so I want to give proper time for people to move away from it. Realistically it will probably remain for a long time (although completely removed from any docs).

As for the functionality you're describing, it was one of the main motivations behind it! I'll have to let this new approach brew for a while, but I'd imagine a simpler hook API like the following would emerge soon:

```lua
require("nvim-lsp-installer").setup {
    hooks = {
        server_installed = function (server_name)
            lspconfig[server_name].setup { }
        end
    }
}
```

mehalter:
Ah awesome @williamboman. I also was able to move to using the `get_installed_servers()` function in my configuration to remedy this deprecation. I am one of the developers for the AstroNvim configuration and we wanted to move our LSP code for our users to use the new APIs. Instead of having a hook when we build the LSP set up we just check which servers are actively installed. Should be a good use case for now. This also makes our configuration much easier for users to manage both language servers installed with `nvim-lsp-installer` and ones that they already have installed on their machine. Thanks so much for the great work!

neevek:
> > Hi, just wondering if there is a migration guide for this breaking change for those using `on_server_ready`.
> 
> Ah I see some people have interpreted this as a breaking change - there are no breaking changes introduced here! Everything will continue to function as before, there's just a new preferred way of setting things up. I see that one of the commit messages was marked as breaking change and made it through in the squashed commit message, so packer is picking that up as a breaking change when updating the plugin version. I'm drafting an announcement with a guide on how to migrate atm, will publish under Discussions when it's ready!
> 
> > Unless I'm missing something, I think you might've forgotten to mention how `lspconfig` would find the server, like read from `server:get_default_options()`, or maybe add the installation root-dir to path.
> 
> I'm not sure that needs to be documented, does it? Just call `.setup()`, then you can keep nvim-lsp-installer out of sight & out of mind. Tbh I'm hoping to either completely get rid of the `Server` class, or alternatively just expose `:install()`, `:is_installed()`, and `:uninstall()` as public methods.

Hi, I am still new to this stuff, but what's the alternative to `server:get_default_options()`? With `on_server_ready`, I `server:get_default_options()` to get the default options and merge them with my custom options, with the new API I don't know how to do that.

williamboman:
> Hi, I am still new to this stuff, but what's the alternative to `server:get_default_options()`? With `on_server_ready`, I use `server:get_default_options()` to get the default options and merge them with my custom options, with the new API I don't know how to do that.

Have you seen https://github.com/williamboman/nvim-lsp-installer/discussions/636? Hopefully it answers your questions, if not send a comment over there instead!

PR #635, opened 2022-04-28
demo: migrate to new way of setting up LSP servers

williamboman:


vuki656:
Just a heads up for anyone still using [nvim-lsp-ts-utils](https://github.com/jose-elias-alvarez/nvim-lsp-ts-utils): it has been put into maintenance mode and the replacement is [typescript.nvim ](https://github.com/jose-elias-alvarez/typescript.nvim)

williamboman:
> The function `.get_installed_servers()` will still be supported in the future?

That function will remain!

bennypowers:
> > The function `.get_installed_servers()` will still be supported in the future?
> 
> That function will remain!

Does that include `server:on_ready(function() ... end)` as in:
```lua
local servers = {
  ['tsserver'] = {
    on_attach = function() ... end,
    root_dir = ...,
    settings = ...
  },
}

-- Loop through the servers listed above.
-- installing each, then if install succeeded,
-- setup the server with the options specified in server_opts,
-- or just use the default options
--
for name, opts in ipairs(servers) do
  opts = opts or {}
  opts.on_attach = opts.on_attach or default_on_attach
  local available, server = lsp_installer_servers.get_server(name)
  if available then
    server:on_ready(function ()
      server:setup(opts)
    end)
    -- Queue the server to be installed.
    if not server:is_installed() then
      server:install()
    end
  end
end
```

PR #637, opened 2022-04-28
feat: add ensure_installed to setup table

williamboman:


williamboman:
Hey @MurdeRM3L0DY! Yes! , see #638, wdyt?

MurdeRM3L0DY:
I think you could also add an `exclude` key for servers users might prefer to install with their system's package manager(clangd for example). Users would be able to configure like so:
```lua
require('nvim-lsp-installer').setup {
  ...
  servers = {
    ensure_installed = { ... } -- or "automatic",
    exclude = { ... },
  },
  ...
}
```

Although, on a second thought. this might make configuration for users who use both nvim-lspconfig together with other lsp plugins awkward(rust-tools for example). So some part of me thinks we shouldn't have the automatic option in the first place since most likely users will only ever install the servers once on a single machine.

williamboman:
Ah, good point!

> So some part of me thinks we shouldn't have the automatic option in the first place since most likely users will only ever install the servers once on a single machine.

This is anecdotal of course but I jump between machines quite a lot, and apart from on my main dev machine I have no idea what's installed, so having something to take care of that for me would help tremendously.

MurdeRM3L0DY:
That's also a fair point. Most of lsp extention plugins call lspconfig setup anyways.

MurdeRM3L0DY:
Alternatively we could install servers automatically when users configure via lspconfig and reserve `ensure_installed` for servers which may or may not be used by lspconfig.

williamboman:
I kept `ensure_installed` as-is, and introduced a `automatic_installation` setting instead. Brain is fried, not sure if naming or docs make any sense. Wdyt?

MurdeRM3L0DY:
Yeahhh, naming is hard :) I can't think of a better name as well.

PR #638, opened 2022-04-28
feat: add automatic_installation option to the setup table

williamboman:
This allows users to automatically install all servers that are setup via lspconfig.

### DIFF
--- a/lua/nvim-lsp-installer/middleware.lua
+++ b/lua/nvim-lsp-installer/middleware.lua
@@ -1,5 +1,6 @@
 local util = require "lspconfig.util"
 local servers = require "nvim-lsp-installer.servers"
+local settings = require "nvim-lsp-installer.settings"
 
 local M = {}
 
@@ -25,6 +26,9 @@ function M.register_lspconfig_hook()
         local ok, server = servers.get_server(config.name)
         if ok then
             merge_in_place(config, server._default_options)
+            if settings.current.automatic_installation and not server:is_installed() then
+                server:install()
+            end
         end
     end)
 end
--- a/README.md
+++ b/README.md
@@ -114,7 +114,8 @@ Example:
 
 ```lua
 require("nvim-lsp-installer").setup({
-    ensure_installed = { "sumneko_lua", "rust_analyzer" },
+    ensure_installed = { "rust_analyzer", "sumneko_lua" },
+    automatic_installation = true,
     ui = {
         icons = {
             server_installed = "✓",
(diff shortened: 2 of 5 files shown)


lua/nvim-lsp-installer/middleware.lua
```diff
@@ -25,6 +26,9 @@ function M.register_lspconfig_hook()
         local ok, server = servers.get_server(config.name)
         if ok then
             merge_in_place(config, server._default_options)
+            if settings.current.automatic_installation and not server:is_installed() then
+                server:install()
```
wookayin, line 30: This is not blocking, isn't it? So installing would take some time for downloading and unpacking, say 10 seconds. But this hook will be executed when lspconfig.setup is called. How do we ensure server is ready and available at this moment?

williamboman: Nope, it's not blocking!

>  How do we ensure server is ready and available at this moment, or lspconfig[...].setup is scheduled to be invoked later after the installation is complete?

We don't, anymore. It's one of the drawbacks, but I'd almost argue that's now a usability issue with lspconfig?

wookayin: Yeah, so my understanding is correct. Actually this was to point out some potential problems in the logical flow rather than just questions --- so even if we do `server:install()`, the installation won't be complete until the actual `lspconfig[..].setup { }` call that immediately follows this `on_setup` hook. Therefore the missing LSP still not work as expected this time (despite `automatic_installation`), with a hope that the missing LSP will work next time, which can be a bit confusing to users. I know there's nothing much we can do here other than this, but at least there could be some notifications or documentation about the timing and behavior.

williamboman: Yeah I want to improve this, will let the new things settle for a while first!

williamboman:
> This is an awesome feature I am looking forward to using it. Is there any way for us to improve this to check the current user's path to see if they already have an language server installed outside of `lsp-installer` and not automatically install those?

Personally I'm not a fan of such a feature, although I see it could be valuable. I guess what's done in https://github.com/williamboman/nvim-lsp-installer/pull/643 could be further extended in the future to allow skipping servers that are already available on PATH (although, what "being available" actually means is up for debate)

mehalter:
@williamboman I looked into this a bit and made an example PR (#645) of what I have in my head. Let me know what you think or if you think it aligns with what you envision. Thanks so much!

PR #642, opened 2022-04-29
Blacklist servers for automatic installation

mxdevmanuel:
Added a setting "avoid_installation" as a list of servers to be avoided in automatic installation. The use case been a user installing an LSP externally but still wanting to use a streamlined config to automatically install all of its other servers except for specific ones.


williamboman:
Hello! Ah yes, that'd make sense too :). See #643 for an alternative approach, wdyt?

mxdevmanuel:
I like the approach but I also think It may be confusing for anyone reading the config, it may looke as a repeated ensure_installed key, although with a little reading of the docs I don't think anyone would confuse.

williamboman:
Superseded by #643!

PR #645, opened 2022-04-29
feat(middleware): add 'ignore_in_path' feature to ignore unnecessary automatic installations

mehalter:
This adds the ability to ignore automatic installations if an language server command is already available on the user's PATH. I currently have it in place to only be regarding the automatic installer. If the user manually installs an LSP then it should still install as they are specifically choosing to install it.

### DIFF
--- a/lua/nvim-lsp-installer/middleware.lua
+++ b/lua/nvim-lsp-installer/middleware.lua
@@ -1,3 +1,4 @@
+local lspconfig = require "lspconfig"
 local util = require "lspconfig.util"
 local servers = require "nvim-lsp-installer.servers"
 local settings = require "nvim-lsp-installer.settings"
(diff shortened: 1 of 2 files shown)


mehalter:
If the user provides the LSP in the `ensure_installed` list it also ignores the check for if the executable is found in the path because it would make sense to respect the user's decision to "ensure the server is installed" with `nvim-lsp-installer`

williamboman (commented):
Thanks! While this would work for a majority of servers, there are servers where this logic would not really work, for example:

- https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/spectral/init.lua#L26
- https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/bicep/init.lua#L30
- https://github.com/williamboman/nvim-lsp-installer/blob/main/lua/nvim-lsp-installer/servers/powershell_es/init.lua#L21

I think for a feature like this to really work, it'd need to be hosted in a different layer that have the necessary capabilities to properly determine whether something is already available or not. I also can't help but feel like this "smartness" would carry a bunch of edge cases (can you tell, I'm not exactly a fan haha :D).

In the past days there's been more changes and new features than I'm comfortable with haha, so I'll let this one sit for a while (the feature itself, not this implementation per se).

I'd love to know more about concrete examples of use cases behind this?

PR #648, opened 2022-05-01
update hls asset file name for version >= 1.7

daschw:
With v1.7.0.0 the asset file names changed for the haskell language server (https://github.com/haskell/haskell-language-server/releases/tag/1.7.0.0). With this change `:LspInstall hls` works again.


williamboman (commented):
Thanks! I haven't verified, but I believe this will only download the source files and unpack them. This won't give you an executable language server. I believe what we need to do is to actually download the assets from the new location

williamboman:
See #651 :)

daschw:
Oh, sorry, you are right. Thanks for fixing!

PR #655, opened 2022-05-01
Add robotframework_ls

AnonymusRaccoon:


williamboman (approved):
Cool! There's a script that generates some code that needs to be ran:

```sh
$ scripts/autogen_metadata.sh
```

I'd be able to do so for you if you were to use a different branch than `main` in your fork (if I were to directly push to the default branch of your fork, GitHub will think I'm partly responsible of the repository and I'll start receiving notifications for it)

williamboman:
Thanks 🎯!

Issue #658, opened 2022-05-02
There are two *DEPRECATED* in Docs.


ryota2357:
### Problem description

The same help tag, \*DEPRECATED\*, is found on lines 378 and 385 of nvim-lsp-installer.txt.
https://github.com/williamboman/nvim-lsp-installer/blob/main/doc/nvim-lsp-installer.txt#L378-L385

I think you need to add a comment as well as the other DEPRECATED tag.

### Neovim version (>= 0.6)

NVIM v0.7.0
Build type: Release
LuaJIT 2.1.0-beta3

### Operating system/version

Darwin MacBook-Pro-2018.local 21.4.0 Darwin Kernel Version 21.4.0: Fri Mar 18 00:45:05 PDT 2022; root:xnu-8020.101.4~15/RELEASE_X86_64 x86_64

### I've recently downloaded the latest plugin version of both nvim-lsp-installer and nvim-lspconfig

- [X] Yes

### Affected language servers

all

### Actual behavior

If you are using [dein.vim](https://github.com/Shougo/dein.vim), dein will warn.

### Expected behavior

no warning

### Healthcheck output

```shell
nvim-lsp-installer: require("nvim-lsp-installer.health").check()
========================================================================
## nvim-lsp-installer report
  - OK: neovim version >= 0.7.0
  - WARNING: **Go**: not available
  - WARNING: **Composer**: not available
  - WARNING: **julia**: not available
  - OK: **tar**: `bsdtar 3.5.1 - libarchive 3.5.1 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.8 `
  - OK: **gzip**: `Apple gzip 353.100.22`
  - OK: **curl**: `curl 7.79.1 (x86_64-apple-darwin21.0) libcurl/7.79.1 (SecureTransport) LibreSSL/3.3.5 zlib/1.2.11 nghttp2/1.45.1`
  - WARNING: **java**: not available
  - WARNING: **javac**: not available
  - OK: **cargo**: `cargo 1.59.0`
  - OK: **wget**: `GNU Wget 1.21.3 built on darwin21.3.0.`
  - OK: **bash**: `GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin21)`
  - OK: **sh**: `Ok`
  - OK: **node**: `v18.0.0`
  - OK: **python3**: `Python 3.10.3`
[...]

PR #660, opened 2022-05-02
docs: add comment to duplicate *DEPRECATED* tag

d4ilyrun:
Hello !

As stated in [this](https://github.com/williamboman/nvim-lsp-installer/issues/658) issue, the documentation contains a duplicated *DEPRECATED* tag which prevents some systems from building the plugins correctly.

I've added a comment to one of the tags saying to configure the servers through `lsp-config` directly (as it had been done for previous deprecated functions).

Thank you for your work and your amazing plugin 😄

### DIFF
--- a/doc/nvim-lsp-installer.txt
+++ b/doc/nvim-lsp-installer.txt
@@ -382,7 +382,7 @@ class: Server
                     nvim-lsp-installer, and not the default options provided
                     by lspconfig.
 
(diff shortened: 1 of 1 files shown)


williamboman:
Ah, woops. Thanks! Fixed in https://github.com/williamboman/nvim-lsp-installer/commit/aee7e760d2970bf3e0e30194d5c26db355894580

PR #667, opened 2022-05-04
fix(hls): install via ghcup install

williamboman:
@ThoNohT @black-desk does this improve things?

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ install _all_ servers are:
 -   valac (and meson & ninja)
 -   rebar3
 -   cargo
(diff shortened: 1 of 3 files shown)


black-desk:
work for me

ThoNohT:
Yeah, this works.

I don't think its much of an issue to require ghcup to be installed. I assume you just call the installed executable and it is not included as a dependency somehow?

If so, it would be good to add it to the *full requirements* in the Insallation section of the README.md. And maybe somewhere to specify that ghcup is required for hls specifically.

Issue #479, opened 2022-02-13
Spawning language server with cmd: `.../nvim/lsp_servers/omnisharp/omnisharp/run` failed. The language server is either not installed, missing from PATH, or not executable.


nyngwang:
### Problem description

After I reinstalled omnisharp via `:LspInstall omnisharp`. It doesn't work because there is no `run` file under `...nvim/lsp_servers/omnisharp/omnisharp/run`, which is a written in the [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#omnisharp).

### Neovim version (>= 0.6)

NVIM v0.7.0-dev+1082-g046950f63
Build type: Release
LuaJIT 2.1.0-beta3

### Operating system/version

macOS 12.2(21D49)

### I've recently downloaded the latest plugin version of both nvim-lsp-installer and nvim-lspconfig

- [X] Yes

### Affected language servers

csharp_ls

### Actual behavior

<img src="https://user-images.githubusercontent.com/24765272/153778545-196e3ea4-aa0f-42b1-aae5-6208df967a46.png" width="800">

### Expected behavior

get it correctly installed by default.

### Healthcheck

```shell
nvim-lsp-installer: require("nvim-lsp-installer.health").check()
========================================================================
## nvim-lsp-installer report
  - OK: neovim version >= 0.6.0
  - WARNING: **Composer**: not available
  - WARNING: **PHP**: not available
  - WARNING: **julia**: not available
  - OK: **tar**: `bsdtar 3.5.1 - libarchive 3.5.1 zlib/1.2.11 liblzma/5.0.5 bz2lib/1.0.8 `
  - OK: **gzip**: `Apple gzip 352.40.6`
  - OK: **curl**: `curl 7.77.0 (x86_64-apple-darwin21.0) libcurl/7.77.0 (SecureTransport) LibreSSL/2.8.3 zlib/1.2.11 nghttp2/1.42.0`
  - OK: **wget**: `GNU Wget 1.21.2 built on darwin21.1.0.`
  - OK: **python3**: `Python 3.9.5`
  - OK: **node**: `v16.8.0`
  - OK: **Ruby**: `ruby 2.6.8p205 (2021-07-07 revision 67951) [universal.x86_64-darwin21]`
  - OK: **Go**: `go version go1.17.6 darwin/amd64`
  - OK: **sh**: `Ok`
[...]

Issue #686, opened 2022-05-11
omnisharp support for Unity using system mono


ModProg:
### Is your feature request related to a problem? Please describe.

It is not possible to use the omnisharp bundled version with unity due to some issues which I don't fully understand myself, but using the system mono works perfectly fine.

This was already discussed in https://github.com/kabouzeid/nvim-lspinstall/issues/127#issuecomment-901588173 with a bunch of references.

### Describe the solution you'd like

Probably something similar as VSCode's `"omnisharp.useGlobalMono": "always"`:

https://github.com/OmniSharp/omnisharp-vscode/wiki/Configuring-Arch-Linux-for-Unity-development

I think this would depend on https://github.com/williamboman/nvim-lsp-installer/issues/479

### Describe potential alternatives you've considered

Add a version of omnisharp using mono, using system mono by default. It is a simple change to the run script, I also currently applied myself to make it work:
https://github.com/OmniSharp/omnisharp-roslyn/issues/1948#issuecomment-696255622

williamboman:
#688 would fix this!

williamboman:
@ModProg Do you think it'd make sense to always use the mono version if `mono` is installed on the system, and only default to the NET6 runtime version when mono isn't available?

ModProg:
I don't know, there must be a reason it normally uses the bundled version in e.g. VSCode. 

I'm not really a c# developer and my only interaction with it is Unity.

There might be advantages to the bundled version that I don't know of.

williamboman:
Sorry just to clarify, what do you mean by the bundled version?

ModProg:
> Sorry just to clarify, what do you mean by the bundled version?

If I understood it correctly omnisharp bundles a mono version? or was that a missunderstanding on my side.

williamboman:
Resolved in #688 and #705!

PR #688, opened 2022-05-11
Add mono version of omnisharp

sassanh:
Fixes https://github.com/williamboman/nvim-lsp-installer/issues/479


sassanh:
Pending https://github.com/neovim/nvim-lspconfig/pull/1898

williamboman (commented):
Cool! I'm however not so sure about adding a new installer for a server that doesn't exist in lspconfig `omnisharp_dotnet`. There's currently a very strong coupling between these, and so this wouldn't be particularly useful as-is.

I'm wondering if we should stick to the existing server, but either

1) Always install both mono and net6 variants, and make it possible to configure which one to use via a setting
2) Make the installer interactive by prompting which variant to install (there's a `std.input()` function for this)

I'm strongly leaning towards 1), for which I'm picturing something like:

```lua
lspconfig.omnisharp.setup {
    use_mono = false,
}

-- and then in omnisharp/init.lua in nvim-lsp-installer
default_options = {
  on_new_config = function (config)
      if config.use_mono then
         config.cmd = { "mono", path.concat { ... }, ... }
      else
         config.cmd = { "dotnet", path.concat { ... }, ... }
      end
  end
}

```

Something like this was also suggested in #686

sassanh:
Yeah I also think 1) is a good option, but maybe eventually we need to decouple server names from lspconfig filenames as lspconfig is not meant to have a configuration for each random setup of a server, it is just best effort configs: https://github.com/neovim/nvim-lspconfig/pull/1898#issuecomment-1124384402

So while I go with option 1) for this PR, we can consider introducing a mapping between server names and filenames in lspconfig so that the names don't need to be the same anymore. What do you think?

sassanh:
@williamboman I updated the PR accordingly, if mono is available globally or if `mono_path` is provided in the config, it will use the mono version, otherwise it will use the dotnet version.

lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -7,55 +7,83 @@ local std = require "nvim-lsp-installer.core.managers.std"
 
 local coalesce, when = functional.coalesce, functional.when
 
+function GenerateCmd(mono_path, root_dir)
```
williamboman, line 10: Without the `local` keyword, it becomes a global!

```suggestion
local function generate_cmd(mono_path, root_dir)
```

lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -7,55 +7,83 @@ local std = require "nvim-lsp-installer.core.managers.std"
...
+  if mono_path == nil and vim.fn.executable("mono") then
+    mono_path = "mono"
+  end
```
williamboman, line 13: I don't see a reason to take a `mono_path` as a setting? Wouldn't it be enough to assume it is on path, and instead let the user decide via a boolean flag, like

```lua
lspconfig.omnisharp.setup {
  use_mono = true
}
```

Or am I missing something here?

williamboman, line 13: I don't see a reason to take a `mono_path` as a setting? Wouldn't it be enough to assume it is on path, and instead let the user decide via a boolean flag, like

```lua
lspconfig.omnisharp.setup {
  use_mono = true
}
```

Or am I missing something here?

lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -7,55 +7,83 @@ local std = require "nvim-lsp-installer.core.managers.std"
...
+              )
+            ),
+            when(
+              platform.is_win,
+              coalesce(
+                when(platform.arch == "x64", "omnisharp-win-x64-net6.0.zip"),
+                when(platform.arch == "arm64", "omnisharp-win-arm64-net6.0.zip")
+              )
+            )
+          ),
```
williamboman, line 70: It's probably fine, but this will make the installer fail if it's unable to find an appropriate net6.0 distribution. We seem to cover pretty much all OS and arch combinations so I think it's fine!

williamboman, line 70: It's probably fine, but this will make the installer fail if it's unable to find an appropriate net6.0 distribution. It might be nice to allow users to continue the installation of Mono, even if net6.0 fails, but we seem to cover pretty much all OS and arch combinations so I think it's fine!

lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -7,55 +7,83 @@ local std = require "nvim-lsp-installer.core.managers.std"
...
+      ctx.fs:mkdir "omnisharp-mono"
+      ctx:chdir("omnisharp-mono", function()
+        github.unzip_release_file({
+          repo = "OmniSharp/omnisharp-roslyn",
+          asset_file = "omnisharp-mono.zip",
+        }).with_receipt()
+      end)
+    end,
+    default_options = {
+      cmd = GenerateCmd(nil, root_dir),
```
williamboman, line 83: Since we're always overwriting the `config.cmd` inside `on_new_config` (this is called by lspconfig whenever it realizes it needs to start the new server), this initial `cmd` value is actually not needed:

```suggestion
```

williamboman, line 83: Since we're always overwriting the `config.cmd` inside `on_new_config` (this is called by lspconfig whenever it realizes it needs to start the new server), this initial `cmd` value is actually not needed:

```suggestion
```

williamboman (commented):
Cool! Just some questions and suggestions. I think we should keep it simple with the mono toggling - let's keep it as a boolean flag and assume `mono` or `dotnet` is on path?

lua/nvim-lsp-installer/servers/omnisharp/init.lua
sassanh: If mono is available I think it is preferred to use the mono version, this is because the dotnet version as far as I understand is a minimal version that is not supporting external libraries, other than that the two are identical as far as I know and the dart version is self sufficient to run.

All that said, if mono is available, we should use it, if it is not available we use the dotnet version. Since other servers are not taking any binary path we can drop `mono_path` altogether and not introduce `use_mono` neither if it sounds good to you.

lua/nvim-lsp-installer/servers/omnisharp/init.lua
sassanh: The credit of this part is not for me, it is just how it was handled already.

sassanh:
@williamboman I changed the commit based on your review comments.

williamboman (approved):
Thanks!! Let's see if people think this makes sense :)

lua/nvim-lsp-installer/servers/omnisharp/init.lua
```diff
@@ -4,9 +4,30 @@ local path = require "nvim-lsp-installer.core.path"
 local functional = require "nvim-lsp-installer.core.functional"
 local github = require "nvim-lsp-installer.core.managers.github"
 local std = require "nvim-lsp-installer.core.managers.std"
+local Result = require "nvim-lsp-installer.core.result"
 
 local coalesce, when = functional.coalesce, functional.when
 
+local generate_cmd = function(root_dir)
+    if vim.fn.executable "mono" then
```
wtfsck, line 12: This fn doesn't return a boolean, it returns an integer so you must check for == 0 or != 0. This currently forces mono even though it's not installed.

PR #693, opened 2022-05-13
Updates Fortran Language Server

gnikit:
switches to `fortls` from `fortran-language-server`.  `fortls` is a continuation of the `fortran-language-server` since it got abandoned ~2y ago. 

The name of the binary was chosen to be the same to allow VS Code, Atom and other editors to be easily backwards compatible without having to rewrite every extension. For that reason `fortls` and `fortran-language-server` cannot be simultaneously installed in a system, since the binary will be overwritten with whichever was installed last.

I also noticed that there is an autogenerated schema file: `fortls.lua` taken from the Fortran Intellisense VS Code extension that should preferably be deleted. That interface was problematic with the original `fortran-language-server` and is completly incompatible with `fortls` and the [Modern Fortran](https://github.com/fortran-lang/vscode-fortran-support) VS Code extension

Full disclosure I am the author of both fortls and Modern Fortran

### DIFF
--- a/lua/nvim-lsp-installer/servers/fortls/init.lua
+++ b/lua/nvim-lsp-installer/servers/fortls/init.lua
@@ -5,9 +5,9 @@ return function(name, root_dir)
     return server.Server:new {
         name = name,
         root_dir = root_dir,
-        homepage = "https://github.com/hansec/fortran-language-server",
+        homepage = "https://github.com/gnikit/fortls",
         languages = { "fortran" },
-        installer = pip3.packages { "fortran-language-server" },
+        installer = pip3.packages { "fortls" },
         default_options = {
             cmd_env = pip3.env(root_dir),
         },


williamboman (approved):
Cool, thanks for the thorough explanation!

williamboman:
> I also noticed that there is an autogenerated schema file: `fortls.lua` taken from the Fortran Intellisense VS Code extension that should preferably be deleted. That interface was problematic with the original `fortran-language-server` and is completly incompatible with `fortls` and the [Modern Fortran](https://github.com/fortran-lang/vscode-fortran-support) VS Code extension

These are sourced from [this gist](https://gist.github.com/williamboman/a01c3ce1884d4b57cc93422e7eae7702). Would it make sense to source the configuration schema from the Modern Fortran VSCode extension instead, or would none of those settings apply to the LSP server itself?

The configuration schema is rendered in the `:LspInstallInfo` window, to aid discovery of server settings.

Screenshot 2022-05-13 at 13 54 30

williamboman:
I've now removed `fortls` mapping from the gist, so the autogenerated code will disappear next time the scheduled GitHub action runs.

gnikit:
Thanks @williamboman much appreciated. As for the schema what I can try and do is export all the `fortls` options from https://github.com/gnikit/fortls/blob/master/fortls/interface.py into a .json that can be imported to editor extensions.

I faced this problem when I redesigned VS Code's extension settings, [see](https://github.com/fortran-lang/vscode-fortran-support/blob/e598472fc547021f1a1f525c4b988f9c0d41f8bf/package.json#L235-L292), but I was lazy and I only enabled a few options from `fortls` in the `package.json` and then I added an `extraArgs` option for all the rest. I will let you know how I get on with this and I will ping you if I have something that works.

FYI `fortls` users can set options through config files so this is not that big of an issue for the end user.

williamboman:
> As for the schema what I can try and do is export all the `fortls` options from https://github.com/gnikit/fortls/blob/master/fortls/interface.py into a .json that can be imported to editor extensions.

That would be amazing ⭐! pylsp is the only server I've seen that [does this currently](https://github.com/python-lsp/python-lsp-server/blob/develop/pylsp/config/schema.json), and it's very much appreciated for non-VSCode users.

PR #729, opened 2022-05-28
refactor(solang): don't download llvm

williamboman:
This is now assumed to be installed on the system instead.


williamboman:
This was reverted, I forgot that solang requires a custom llvm build

PR #807, opened 2022-06-17
fix lombok not being recognized.

jarrin-p:
I read that the `-javaagent:...` needs to come before the `-jar` flag. Changing the order made `Lombok` work again.


williamboman (approved):
Ah, thanks!

wtfsck:
This PR broke `jdtls`: `Client 1 quit with exit code 1 and signal 0`, using `openjdk 18-ea 2022-03-22`.

The fix is to remove the empty `""` from the command line to get it working again.

```lua
use_lombok_agent and ("-javaagent:" .. lombok) or "",
```

williamboman:
Ah, I wasn't sure whether it'd accept empty strings, I just assumed it would 🤦‍♂️. Fixed in https://github.com/williamboman/nvim-lsp-installer/pull/814

Issue #763, opened 2022-06-08
jdtls does not work


aman7000:
### Problem description

Java lsp does not work

### Why do you think this is an issue with nvim-lsp-installer?

getting error on java file, instead of lsp loading. Error:-
Client 1 quit with exit code 13 and signal 0

### Neovim version (>= 0.7)

NVIM 0.7.0

### Operating system/version

Mac

### I've manually reviewed the Nvim LPS client log (`:LspLog`) to find potential errors

- [ ] Yes

### I've recently downloaded the latest plugin version of both nvim-lsp-installer and nvim-lspconfig

- [X] Yes

### Affected language servers

java

### Steps to reproduce

1. Install java server with :LspInstall java
2. Open java file to see the error

### Actual behavior

error comes:
Client 1 quit with exit code 1 and signal 0

### Expected behavior

java Lsp should work

### LspInfo

```Text
Language client log: /Users/amansingh/.cache/nvim/lsp.log
 Detected filetype:   java
 
 0 client(s) attached to this buffer: 
 
 Other clients that match the filetype: java
 
 Config: jdtls
 	filetypes:         java
 	root directory:    /Users/amansingh/workspace/java/nbfc
 	cmd:               /Library/Java/JavaVirtualMachines/zulu-8.jdk/Contents/home/jre/bin/java -Declipse.application=org.eclipse.jdt.ls.core.id1 -Dosgi.bundles.defaultStartLevel=4 -Declipse.product=org.eclipse.jdt.ls.core.product -Dlog.protocol=true -Dlog.level=ALL -Xms1g -Xmx2G -javaagent:/Users/amansingh/.local/share/nvim/lsp_servers/jdtls/lombok.jar --add-modules=ALL-SYSTEM --add-opens java.base/java.util=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED -jar /Users/amansingh/.local/share/nvim/lsp_servers/jdtls/plugins/org.eclipse.equinox.launcher_1.6.400.v20210924-0641.jar -configuration /Users/amansingh/.local/share/nvim/lsp_servers/jdtls/config_mac -data /Users/amansingh/workspace/nbfc
 	cmd is executable: false (checked by nvim-lsp-installer)
 	                   Make sure you have set up nvim-lsp-installer (:h nvim-lsp-installer-quickstart)
 	autostart:         true
 	custom handlers:   language/status, textDocument/rename, workspace/applyEdit, textDocument/codeAction
[...]

PR #808, opened 2022-06-17
allow user to define `JAVA_HOME` in `...jdtls.setup{}`

jarrin-p:
This is mostly a convenience I'm using that makes dealing with issues like #763 easier, figured I'd share.

I use a few different versions of java. Recently it seems like `jdtls` has a new dependency of Java version >= 17. Since `jdtls` doesn't really need to change versions afaik, I found it more portable to be able to pass a `JAVA_HOME` path during `.setup{}`, i.e.

```
require('lspconfig').jdtls.setup {
    ...
    java_home = '/path/to/jdk/17',
    ...
}
```

Opening this as a draft since I'm not familiar with testing this.

### DIFF
--- a/lua/nvim-lsp-installer/servers/jdtls/init.lua
+++ b/lua/nvim-lsp-installer/servers/jdtls/init.lua
@@ -5,14 +5,17 @@ local _ = require "nvim-lsp-installer.core.functional"
 local installer = require "nvim-lsp-installer.core.installer"
 local eclipse = require "nvim-lsp-installer.core.clients.eclipse"
 local std = require "nvim-lsp-installer.core.managers.std"
+local settings = require "nvim-lsp-installer.settings"
 
 return function(name, root_dir)
     ---@param workspace_root string
     ---@param workspace_path string|nil @The path to the server instance's current workspace. Can be nil when running in single file mode.
     ---@param vmargs string[]
     ---@param use_lombok_agent boolean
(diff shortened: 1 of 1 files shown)


williamboman:
Hello! I guess you could effectively achieve the same thing by passing it as an environment, like so:

```lua
lspconfig.jdtls.setup {
  cmd_env = {
    JAVA_HOME = "/path/to/jdk17"
  }
}
```

Also, recently there was a wrapper script [merged upstream ](https://github.com/eclipse/eclipse.jdt.ls/pull/2005). I'm planning on moving towards running that executable to standardize things. The `cmd` would probably be removed altogether from nvim-lsp-installer and instead use the [one governed by lspconfig](https://github.com/neovim/nvim-lspconfig/blob/f6e369e9e7feb0d8792558dc3ecf1616b19e6b2f/lua/lspconfig/server_configurations/jdtls.lua).

jarrin-p:
Ah! Thank you, that's much better.

williamboman:
> Thank you! that's much better. Where can I find what arguments can be passed to .setup{}s usually?

That's in [lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md)!

There are some servers in nvim-lsp-installer that take extra arguments - these servers have their own README page [linked here](https://github.com/williamboman/nvim-lsp-installer#available-lsps)

PR #812, opened 2022-06-18
docs: add missing rnix in Available LSPs

ocfox:
rnix has been add at #408, but readme has not updated.


williamboman (approved):
Thanks ⭐️!

PR #813, opened 2022-06-18
fix the taplo with the upstream

Decodetalkers:
Because the lsp is not the default feature in taplo, so need to add the feature


williamboman (approved):
Ah, thanks! Seems like it was changed [here](https://github.com/tamasfe/taplo/commit/78856eb767a1c1b95d26d77f59b0cafa3d11abf5). Not sure if it was even intentional

PR #826, opened 2022-06-22
Configs passed by lspconfig.setup{...} would be overwritten.

qianwch:
FIX1: configs passed by require'lspconfig'.xxxx.setup{...} would be overwritten by the on_setup
FIX2: jdtls on_new_config hook functions could overwrite the config.cmd passed by setup{}.


williamboman:
This project is no longer maintained, see https://github.com/williamboman/mason.nvim!

PR #828, opened 2022-06-22
fix(gunzip): win_extract

### LINKED ISSUE #827: Windows: Installation failed, rust_analyzer, unable to unpack gz file
https://github.com/williamboman/nvim-lsp-installer/blob/5a0bf929eefad999e92963ee227c23a804cf0788/lua/nvim-lsp-installer/core/managers/std/init.lua#L159

# Issue
During the installation of the rust_analyzer on the windows platform occurs an error caused by unpacking the corresponding gz file.

# Solution
Replacing the function for win platform with the code from the unix platform function.
This is possible because gunzip is an inbuilt tool in window >= 10

DariusCorvus:
extracted the unix 'gunzip' function into a local function named 'gunzip'
assigned the 'gunzip' to the platform unix and win

win_extract is obsolete in windows >= 10.

fixes #827

### DIFF
--- a/lua/nvim-lsp-installer/core/managers/std/init.lua
+++ b/lua/nvim-lsp-installer/core/managers/std/init.lua
@@ -150,14 +150,14 @@ end
 ---@async
 ---@param file string
 function M.gunzip(file)
+    local function gunzip()
+        local ctx = installer.context()
+        ctx.spawn.gzip { "-d", file }
+    end
+    
     platform.when {
-        unix = function()
-            local ctx = installer.context()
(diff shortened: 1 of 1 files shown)


williamboman:
This project is no longer maintained, see https://github.com/williamboman/mason.nvim. Feel free to port this change there if you'd like, however I'd like to get to the bottom of whether gunzip really comes with stock Windows and is not a cygwin/mingw tool (https://github.com/williamboman/nvim-lsp-installer/issues/827#issuecomment-1163518279)

PR #830, opened 2022-06-22
docs: remove bad packer example

williamboman:


anjannath:
thanks, hit this in the morning after tinkering with my config late last night

PR #838, opened 2022-06-28
refactor(ui): use nvim_set_hl for setting highlight groups

smjonas:
Saw this while looking at `display.lua` and thought I'd help out :) I learned a lot from your UI / window code for my plugin by the way, so thank you for that!


williamboman (approved):
That's awesome, thanks!! I pushed a few commit that fixed some broken stuff. I'm glad to hear it's been helpful for you elsewhere :). I think it's still largely a hack haha, I wanted something fully state-driven with a component-ish layout approach to allow for quick iteration and prototyping - but it turns out it actually performs pretty decently so I never bothered revisiting it!

